Issue #31, opened 2014-09-05
Use pecl/uuid for v1 and v4 UUIDs, if it is available


ramsey:

ramsey:
Implemented in #42.

PR #42, opened 2014-12-17
Add PECL based UUID factory for UUID1/4

aztech-dev:
As suggested in #31 

Defaults to using PECL when available, fallsback to PHP factory when extension is not available, or when it is not possible to conform to the interface (parameters declared in uuid1(...) interface are not available via PECL.

Design choices (decorator instead of composing the existing factory) may be questionnable, but it gets the job done transparently, and honestly, laziness.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ php:
 before_script:
   - composer self-update
   - composer install --dev --prefer-source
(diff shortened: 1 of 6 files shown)


ramsey:
No need to squash. I'm merging it in now. Thanks a bunch!

aztech-dev:
You're welcome :)

PR #45, opened 2015-01-07
"passthru" replacement for "backtick operator" / getNodeFromSystem once per runtime 

yberkholz:
- change method "getIfconfig()" to use "passthru()" and an output buffer instead of the "backtick operator" which may be disabled for security reasons
- change  method "getNodeFromSystem()" to process detection only once per runtime - because of great performance impact
(Test: generate 100 UUIDs[v1] before ~21s / after ~200ms)

### DIFF
--- a/src/Uuid.php
+++ b/src/Uuid.php
@@ -1134,20 +1134,21 @@ final class Uuid
      */
     protected static function getIfconfig()
     {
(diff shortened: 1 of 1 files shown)


ramsey:
Looks great. Thanks for the pull request!

PR #46, opened 2015-01-07
Update .gitignore

GrahamCampbell:
This stuff belongs in your system .gitignore file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,4 @@
-.idea
 *.phar
-.DS_Store
 build
 composer.lock
 docs


GrahamCampbell:
Ping @ramsey.

ramsey:
@GrahamCampbell Do these hurt anything by remaining in the project `.gitignore`?

GrahamCampbell:
Yeh and no. While not breaking anything here, they still encourage people to do this bad practice in their own projects.

ramsey:
I don't understand why it's a bad practice. Can you elaborate?

GrahamCampbell:
http://www.programblings.com/2008/10/22/git-global-ignores/

ramsey:
Thanks for the link, but that post seems to imply that, when working in teams, having these ignores in the project `.gitignore` file is still helpful:

> When I create a new repository on which people may actually contribute, I’ll still create a proper ignore file, however. Otherwise I’d convey the message that I consider contributors as slaves who only deserve the boring work of creating ignore files. Since I’m pure of heart, that’s not how I roll.

Since multiple people work on this project, I think it's helpful to keep a robust `.gitignore` file for the project, since I can't assure that all developers will put the files in their global ignore file. Plus, I end up cloning this repo to multiple places, so I personally have to keep my global ignore files up-to-date on multiple systems. IMO, it's better for the project to keep track of the files it doesn't want to be committed.

GrahamCampbell:
@ramsey I don't agree. If people see files being staged by git that they don't expect to see, they'll probably check the docs and realize they're meant to be using a global ignore file.

ramsey:
@GrahamCampbell I don't see the harm in leaving these values in `.gitignore`, so I'm going to leave them in place. I appreciate your interest in this library, and I hope you'll continue to use and support it through future pull requests. Thanks!

PR #49, opened 2015-03-09
getInteger should suggest getHex

francescolaffi:
getInteger is suggesting getMostSignificantBitsHex if the math library is not available


ramsey:
Great catch. Thanks!

PR #57, opened 2015-05-07
Updated readme to show correct composer require command.

chrislentz:
The old composer command pulled down the old "rhumsaa/uuid" version which didn't match the documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,7 +93,7 @@ the following command to install the package and add it as a requirement to
 `composer.json`:
 
 ```bash
-composer require ramsey/uuid
+composer require ramsey/uuid "3.0.*@dev"
 ```
 
 


ramsey:
I don't feel the master branch is in a position to encourage others to start using it yet, which is why the composer command remains where it is. I hope to get to a point where I can release an alpha version of 3.0 in a few weeks, but right now, it's still a bit of a moving target.

Thanks so much for your interest and willingness to contribute!

ramsey:
I wanted to follow up to point out that @arvidbjorkstrom in #59 reminded me that I could set the default GitHub branch to `2.8`, so that's what I've done in the meantime. It should help resolve any confusion while `master` is still in a pre-alpha 3.0 stage.

Issue #56, opened 2015-05-06
Performance drop


jaymecd:
Hi there,

Today I've compared UUID generations with 2.8 vs 3.0 and noticed ~2.5x performance drop.

Pls check gist https://gist.github.com/jaymecd/7dbac8908bacf54b6db6 for more details.

jaymecd:
@ramsey gist updated

ramsey:
Thanks! I've got it loaded up in MacCallGrind now. I'll dig through this and report back what we'll do to proceed.

aztech-dev:
I left a comment on the gist, I **may** have an idea of whats going on

aztech-dev:
After having done some tests, it turns out the root of the issue is that the uuid_create function returns a string UUID which then needs to be parsed into a Uuid instance, and that part is expensive. Not sure how to solve that issue without changing the fundamental premise of the lib -- ie, the Uuid::uuid**x**() start returning string's instead of Uuid instances.

ramsey:
Is it 2.5x slower than Rhumsaa\Uuid because it's using `uuid_create()` **_and_** parsing it into a Ramsey\Uuid object?

aztech-dev:
Exactly. 

My base benchmark results are:
```
melody run test-benchmark.php 
PECL | 0.0410 sec/10000 | 0.0000041 sec/one
RHUMSAA | 0.1500 sec/10000 | 0.0000150 sec/one
RAMSEY | 0.3280 sec/10000 | 0.0000328 sec/one
```

I tested the following changes:

- Return the uuid_create result without parsing => ~ 0.06 seconds
- Remove the PECL factory => ~0.20 seconds (vs. 0.15 seconds on 2.8.x, I assume the overhead is due to the decomposition into various classes)
- Remove the if branch => ~0.35 seconds

aztech-dev:
After a bit of pondering, I came to the following conclusion: most use cases for UUID are all about generating a UUID string (ie. database primary keys etc). A potential workaround for the performance issue would be to add a new Uuid class that stores its field data as a string, and only tries to parse it into fields when necessary. That would give a significant performance boost for most scenarios using the PECL lib. Thoughts ?

ramsey:
It sounds like using the PECL library isn't buying us any efficiencies, as I suspected it would. The primary reason to use it was because it uses `/dev/urandom`, if available.

ramsey:
[...]

PR #58, opened 2015-05-11
Lazy uuid parsing

aztech-dev:
Improves performance when generating UUID's via the PECL UUID extension by delaying parsing of the uuid string until necessary.

Refs #56

### DIFF
--- a/src/LazyUuid.php
+++ b/src/LazyUuid.php
@@ -25,6 +25,7 @@ class LazyUuid extends Uuid
         }
 
         $this->fields = $this->codec->decode($this->uuidString)->getFieldsHex();
+        $this->isParsed = true;
     }
 
     public function compareTo(UuidInterface $uuid)
--- a/src/LazyUuid.php
+++ b/src/LazyUuid.php
@@ -1,0 +1,244 @@
+<?php
+
+namespace Ramsey\Uuid;
+
+use Ramsey\Uuid\Converter\NumberConverterInterface;
+
+class LazyUuid extends Uuid
+{
+
+    private $isParsed = false;
+
+    private $uuidString;
+
+    public function __construct($uuidString, NumberConverterInterface $numberConverter, CodecInterface $codec)
+    {
+        $this->uuidString = $uuidString;
+
+        parent::__construct(array(), $numberConverter, $codec);
+    }
+
+    private function parse()
+    {
+        if ($this->isParsed) {
+            return;
+        }
+
+        $this->fields = $this->codec->decode($this->uuidString)->getFieldsHex();
+    }
+
+    public function compareTo(UuidInterface $uuid)
+    {
+        $this->parse();
+
+        return parent::compareTo($uuid);
+    }
+
+    public function getBytes()
+    {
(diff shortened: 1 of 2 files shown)


jaymecd:
@aztech-dev, not sure if it's a good idea to create another class `LazyUuid`, which overwrites majority methods, as it may bring doubts for implementation use cases and especially when developer expects `LazyUuid` instance in his method by mistake.

Think it would be fine to implement this patch inside `Uuid` class directly.

aztech-dev:
Not too sure also, but not convinced Uuid is the right place either. Can you clarify your point about doubts for implementation ? I'm not sure I understand what you mean... Thanks :)

src/LazyUuid.php
```diff
@@ -0,0 +1,244 @@
...
+        parent::__construct(array(), $numberConverter, $codec);
+    }
+
+    private function parse()
+    {
+        if ($this->isParsed) {
+            return;
+        }
+
+        $this->fields = $this->codec->decode($this->uuidString)->getFieldsHex();
```
aztech-dev: Missing assignment : $this->isParsed = true

jaymecd:
Ah, my bad :) Completely forgot about `UuidInterface`. 
Anyway, my explanation - normally you rely on a contract (interface), but `Uuid` is a class-as-a-contract (**before 3.0**). And now we bring `LazyUuid` class-as-a-contract.
~~Thats my problem~~
U could skip this comment, hard Monday :)

aztech-dev:
Ok :) Good thing I asked... but LazyUuid extends Uuid (most people are type-hinted against Uuid, not UuidInterface, so I felt it was safer) anyways, so it should not matter.

ramsey:
I'm also concerned about `LazyUuid`. Should we lazy-parse always for the performance gain, or are we only getting the performance benefit of lazy parsing when we're using `PeclUuid`?

aztech-dev:
I originally went for the 'always lazy parse' approach by returning a LazyUuid the `PeclUuidFactory::fromString()`, instead of returning a LazyUuid from the `uuid1()` and `uuid4()` methods. However, some unit tests for `fromString()` were failing since you can get by with passing an invalid/malformed string, so I did not keep the behavior. I'll try to push a smoother implementation in a bit.

ramsey:
I've done a bunch of merging around and making changes, so this is out of sync with master. You may want to rebase or submit a new PR.

ramsey:
@aztech-dev I'm getting ready to release a 3.0.0-alpha1. Is there any chance you can take a look and rebase this against master this week? If not, I can give it a try.

aztech-dev:
Sure, consider it done

aztech-dev:
I've rebased and fixed the (almost) minor issue that was laying around. However, I'm still not too sure about how to handle the lazy parsing. One hand, it could be nice to always lazy parse, on the other, lazy parsing skips validation of given string. Thoughts ?

aztech-dev:
I'm closing this in favor of #68 which feels like a better/cleaner option than this one.

PR #60, opened 2015-05-14
composer.phar require "rhumsaa/uuid=~2.8" is not proper

suzuken:
Then running

    composer.phar require "rhumsaa/uuid=~2.8"

said:

    Package rhumsaa/uuid is abandoned, you should avoid using it. Use ramsey/uuid instead.

so should change it to 

    composer.phar require "ramsey/uuid=~2.8"


ramsey:
Thanks!

PR #68, opened 2015-07-23
Random and time generators for pecl-uuid

ramsey:
This isn't the most elegant solution for abstracting time generation into a `TimeGeneratorInterface` family of classes, but it works as a proof-of-concept for what I'd like to do.

Please feel free to submit pull requests to my `pecl-generator` branch to make this better.

### DIFF
--- a/src/BinaryUtils.php
+++ b/src/BinaryUtils.php
@@ -31,6 +31,4 @@ class BinaryUtils
 
         return $timeHi;
     }
-
-
 }
--- a/src/BinaryUtils.php
+++ b/src/BinaryUtils.php
@@ -1,0 +1,36 @@
+<?php
+
+namespace Ramsey\Uuid;
+
+class BinaryUtils
+{
+    /**
+     * @param $clockSeqHi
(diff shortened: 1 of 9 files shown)


ramsey:
@aztech-dev, what are your thoughts on the `TimeGeneratorInterface::generate()` method signature? I don't feel great about it, but it's what I had to do in order to pull out the default time generator code and put it into `DefaultTimeGenerator`. Any thoughts on how to better separate the dependencies?

aztech-dev:
Yup after a closer look :)

I've issued a P/R against your branch with the changes I'd implement, but it boils down to removing UuidFactory from the signature and injecting the exact dependencies at construction in the DefaultTimeGenerator implementation.

I've had to remove some setters from UuidFactory as a result, so it's now required to build a new factory if you ever need to change the node provider, time provider, or time converter, but the methods weren't used at all except in the tests.

ramsey:
Thanks to @aztech-dev, this is no longer a PoC.

PR #69, opened 2015-07-23
Refactor TimeGeneratorInterface

aztech-dev:
- Remove UuidFactory from signature
- Remove setters on UuidFactory to prevent inconsistencies (a new feature set and factory should be used instead)
- Update tests to reflect those changes

### DIFF
--- a/src/FeatureSet.php
+++ b/src/FeatureSet.php
@@ -29,6 +29,7 @@ use Ramsey\Uuid\Codec\GuidStringCodec;
 use Ramsey\Uuid\Builder\DegradedUuidBuilder;
 use Ramsey\Uuid\Generator\RandomGeneratorFactory;
 use Ramsey\Uuid\Generator\TimeGeneratorFactory;
+use Ramsey\Uuid\Provider\TimeProviderInterface;
 
 /**
  * Detects and exposes available features in current environment (32 or 64 bit, available dependencies...)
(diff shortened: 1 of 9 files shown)


src/FeatureSet.php
```diff
@@ -121,6 +123,12 @@ public function getTimeProvider()
         return $this->timeProvider;
     }
 
+    public function setTimeProvider(TimeProviderInterface $timeProvider)
+    {
+        $this->timeProvider = $timeProvider;
+        $this->timeGenerator = $this->buildTimeGenerator();
```
ramsey: You're calling this as the last line of the constructor, too. Does it also need to be here?

aztech-dev: Yup, due https://github.com/ramsey/uuid/pull/69/files#diff-b8cb7bbf7ebc08e740d40433bd79da3aR169. Proper fix is to instead call `$this->setTimeProvider(new SystemTimeProvider())` in the ctor instead. I probably should explicitely pass the required dependencies to the TimeGeneratorFactory instead of passing the FeatureSet instance, it would make the dependencies more explicit.

I'll update my P/R in a bit.

ramsey:
Awesome work, btw!

src/UuidFactory.php
```diff
@@ -55,7 +55,7 @@ class UuidFactory implements UuidFactoryInterface
      */
     private $timeGenerator = null;
 
-    /**
+    /*
```
GrahamCampbell: ?

ramsey: I wasn't sure why this happened either, but I fixed it in https://github.com/ramsey/uuid/commit/d88a941b8c7b7618fcf1f3c41eaa0deb14cadfea. :-)

aztech-dev: Must have been me... sorry, choppy fingers :D

PR #70, opened 2015-07-28
Time generator improvements

aztech-dev:
- Add option to enable PeclUuidTimeGenerator via FeatureSet
- Enable use of custom TimeGenerator implementations
  - Removes now unnecessary `timeConverter` and `timeProvider`
    properties, setters, and getters in both FeatureSet and
    UuidFactory as those are now exclusively used by the default
    TimeGenerator
  - Adds a `setTimeGenerator` method on UuidFactory to override the
    default time generator

### DIFF
--- a/src/FeatureSet.php
+++ b/src/FeatureSet.php
@@ -14,6 +14,7 @@
 
 namespace Ramsey\Uuid;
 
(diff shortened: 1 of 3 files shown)


ramsey:
:+1:

PR #81, opened 2015-08-28
Better random number generator for UUID v4

bdelespierre:
When openssl isn't available, `mt_rand` doesn't provide enought entropy, i.e. it is not a not cryptographically strong random number generator (as stated here http://php.net/manual/en/function.mt-rand.php).

The @ircmaxell random number generator (https://github.com/ircmaxell/RandomLib) always tend to provide better entropy by using multiple random sources, using it decreases drastically the risk of collision when we want to generate a large amount of random UUID.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,13 +29,15 @@
         "phpunit/phpunit": "~4.1",
         "satooshi/php-coveralls": "~0.6",
         "squizlabs/php_codesniffer": "^2.3",
-        "jakub-onderka/php-parallel-lint": "^0.9.0"
+        "jakub-onderka/php-parallel-lint": "^0.9.0",
+        "ircmaxell/random-lib": "^1.1"
     },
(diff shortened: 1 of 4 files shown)


ramsey:
Thank you for your interest in this library and your time in putting together this contribution!

Please take a look at version `3.0.0-alpha3`. The 3.0 series provides this ability. You may use it like this:

``` php
$uuidFactory = new \Ramsey\Uuid\UuidFactory();
$uuidFactory->setRandomGenerator(new \Ramsey\Uuid\Generator\RandomLibAdapter());
\Ramsey\Uuid\Uuid::setFactory($uuidFactory);

$uuid = \Ramsey\Uuid\Uuid::uuid4();
```

I appreciate your time and effort, but I won't be able to accept this PR, since the functionality already exists in the next version.

bdelespierre:
Understandable :+1:

PR #82, opened 2015-08-31
[2.8] Remove rhumsaa from composer.json

frederikbosch:
Just did a `composer show -i` and saw a big warning NO LONGER MAINTAINED on the ramsay/uuid package. This confused me because I was sure I switched months ago. I found out that there are still references to the old package inside the composer.json.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,9 +1,9 @@
 {
-    "name": "rhumsaa/uuid",
-    "description": "NO LONGER MAINTAINED. Use ramsey/uuid instead. A PHP 5.3+ library for generating RFC 4122 version 1, 3, 4, and 5 universally unique identifiers (UUID).",
+    "name": "ramsey/uuid",
+    "description": "A PHP 5.3+ library for generating RFC 4122 version 1, 3, 4, and 5 universally unique identifiers (UUID).",
     "type": "library",
     "keywords": ["uuid", "identifier", "guid"],
-    "homepage": "https://github.com/ramsey/rhumsaa-uuid",
+    "homepage": "https://github.com/ramsey/uuid",
     "license": "MIT",
     "authors": [
         {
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
What?

frederikbosch:
@GrahamCampbell If you are using `ramsay/uuid` in a project. And you are running `composer show -i` within the project context, you will see.

```ramsey/uuid                             2.8.2   NO LONGER MAINTAINED. Use ramsey/uuid instead. A PHP 5.3+ library for generating RFC 4122 version 1, 3, 4, and 5 universally unique identifiers (UUID).```

If you look at my PR, you will find out that I removed that notification, changed the issue/source urls of the project to the updated one. And also the name of the project contained the old name (`rhumsaa/uuid`).

GrahamCampbell:
That's because 2.8 isn't maintained.

GrahamCampbell:
You want ~3.0.

frederikbosch:
Do you? Those are all alphas.

GrahamCampbell:
Yeh, well, I guess 2.8 was just marked as old slightly to early. That was probably intentional to make people aware of 3.0.

frederikbosch:
@GrahamCampbell In that case, I will still leave this PR open. It is up to the owner to decide. In my view, I think it is odd to ask people to switch to a new library location but still have a notification in the latest stable release of the new location that the library is not maintained any longer.

frederikbosch:
Moreover, I do not see a notification anywhere that I should be switching to 3.0.

* 2.8 is the default branch on Github
* That branch does not tell me to upgrade to 3.0
* The [Packagist](https://packagist.org/packages/ramsey/uuid) page explicitly tells me to use 2.8.

If the owner does not agree, then at least this PR might encourage to update communication on the versions.

ramsey:
3.0 is still alpha. It shouldn't be much longer before I have a beta and final out. It's just taking me a while to add all the docblocks and get it polished up. Otherwise, it's essentially done.

The entire switch to `ramsey/uuid` and `Ramsey\Uuid` has been one big confusing mess. My advice is to never change package names or namespaces. Otherwise, everyone gets confused. :-)

frederikbosch:
@ramsey I can imagine. Thanks for the great work on this lib any way!

frederikbosch:
And for the merge!

PR #83, opened 2015-08-31
[3.0] Removed rhumsaa locations from composer.json

frederikbosch:
Changed old support location to new support location.


ramsey:
Good catch. It was probably a merge up from `2.8` that changed this file in `master`.

PR #87, opened 2015-10-05
Cleanup: Update and add missing docblocks, remove unused imports

emirb:
Just a bit of house keeping for the sake of code cleanliness.


ramsey:
Thanks for catching these!

PR #89, opened 2015-10-08
Modify Travis CI builds to run tests on multiple CPU architectures

ramsey:
The ramsey/uuid library has requirements to run on 32-bit and 64-bit systems, as well as little-endian and big-endian systems. Travis CI only provides hardware for testing 64-bit, little-endian CPU architectures, but they recently announced support for Docker[^1].

After some research, I found it is possible to use [QEMU][] with [chroot][] jails on Travis CI to emulate other CPU architectures and run tests through these chroot jails[^2]. Unfortunately, it takes an awful long time to set up a chroot jail, build PHP within it, and run the tests. It wasn't reasonable to expect to do this on every single push to the ramsey/uuid repository. This is where Docker comes in.

I was able to create chroot jails for each architecture and version of PHP combination I need to test and then turn these into Docker images, which Travis is able to download and run fairly quickly. The instances on which Travis CI is running their Docker services do not currently support phpenv or php-build, so I needed to masquerade as a Python project and specify all PHP versions and CPU architectures in environment variables in my `.travis.yml` configuration. I run `composer install` and report to Coveralls outside of the Docker images, but all tests are run within the Docker images.

The Docker image I created for PHP 7.0.0RC4 on MIPS architecture produces segfaults when attempting to build PHP with Phar. PHP builds okay without Phar, but then it produces segfaults when running any PHP script on the image, so I'm leaving out 7.0.0RC4 on MIPS for now.

The Docker images for use with ramsey/uuid tests are available here: <https://hub.docker.com/r/benramsey/ramsey-uuid/>.

[^1]: "Using Docker on Travis CI," <http://blog.travis-ci.com/2015-08-19-using-docker-on-travis-ci/>
[^2]: "Running Travis CI tests on ARM architecture," <https://www.tomaz.me/2013/12/02/running-travis-ci-tests-on-arm.html>

[qemu]: http://wiki.qemu.org/Main_Page
[chroot]: https://en.wikipedia.org/wiki/Chroot

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -5,3 +5,4 @@
 /tests            export-ignore
 /phpunit.xml.dist export-ignore
 /phpcs.xml        export-ignore
+/util             export-ignore
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
-.idea/
 *.phar
 .DS_Store
+.idea/
+.vagrant/
 build
 composer.lock
 docs
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,30 +1,35 @@
-language: php
+# Fake out Travis CI, since PHP isn't supported when using their Docker services
+language: python
 
-php:
-  - 5.4
-  - 5.5
-  - 5.6
-  - 7.0
-  - hhvm
+env:
+  - PHP_VERSION=5.4.45 ARCH=x86_64
+  - PHP_VERSION=5.4.45 ARCH=mips
+  - PHP_VERSION=5.5.30 ARCH=x86_64
+  - PHP_VERSION=5.5.30 ARCH=mips
+  - PHP_VERSION=5.6.14 ARCH=x86_64
+  - PHP_VERSION=5.6.14 ARCH=mips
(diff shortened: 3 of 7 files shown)


ramsey:
@aztech-dev, Would you mind taking a look at this for review? It doesn't change the library in any way, but I want to make sure I'm not forgetting some aspect of the tests that needs to be covered by the new Docker images. The script I used to create the Docker images is in this PR.

ramsey:
For reference, here are the results of this pull request's build on Travis CI:

https://travis-ci.org/ramsey/uuid/builds/84233504

aztech-dev:
If you insist... but gotta be honest, your bash-fu is way above mine.

ramsey:
I learned a lot while doing this. :-)

ramsey:
Going to go ahead and merge this in.

aztech-dev:
Yeah, sorry I didn't get any time to review but I've been swamped with work these days...

ramsey:
No worries. :-)

PR #92, opened 2015-10-25
Fixes #91: Made Uuid's serializable.

### LINKED ISSUE #91: Serializable Uuid's.
Hi Ben,

I've been using Uuid's instead of incremental ID's for some time now. In some situations I've had to implement the Serializable interface due to output corruption when using regular serialization. An example where this happens is the when an object is stored in native PHP sessions. Would you be open to a PR which makes the Uuid class serializable? My initial thought would be to implement it like this:

```php
public function serialize()
{
    return $this->toString();
}

public function unserialize($serialized)
{
    $uuid = self::fromString($serialized);
    $this->codec = $uuid->codec;
    $this->converter = $uuid->converter;
    $this->fields = $uuid->fields;
}
```

Like the JsonSerializable, the interface could then be added to the UuidInterface. 

Let me know what you think, I'll happily submit a PR with tests.

frankdejonge:
As discussed.

### DIFF
--- a/src/Uuid.php
+++ b/src/Uuid.php
@@ -187,6 +187,32 @@ class Uuid implements UuidInterface
         return $this->toString();
     }
 
+    /**
+     * Converts this UUID object to a string when the object is serialized
+     * with `serialize()`
+     *
+     * @return string
+     * @link http://php.net/manual/en/class.serializable.php
+     */
+    public function serialize()
+    {
+        return $this->toString();
+    }
+
+    /**
+     * Re-constructs the object from its serialized form.
+     *
+     * @param string $serialized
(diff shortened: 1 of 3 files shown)


frankdejonge:
@ramsey those unit test take forever. Don't know what's up with that, but this is it, and it's got the coverage to prove it.

ramsey:
They run slow (especially if you have Xdebug enabled) because of 67,000+ assertions to ensure the math is all correct and will generate the same UUIDs for the same random or time values as in other programming languages. :-)

Thanks for this!

frankdejonge:
@ramsey ahhh, that makes sense. Thanks for the quick merge :+1:

Issue #90, opened 2015-10-21
create DateTime with nanoseconds


thomasmuehlschlegel:
The uuid1 is using nanoseconds, but the UuidInterface returns a DateTime object which does not include the nanoseconds. 

DateTime can handle it if you adjust the initialization:

```
\DateTime::createFromFormat('U.u', sprintf('%.f', ($this->getTimestamp() - 0x01b21dd213814000) / 1e7))
```

ramsey:
Unfortunately, this is failing our tests, since the `U` option does not support negative timestamps. I'm still looking into it, though.

ramsey:
I spoke with Derick about it, and he has already fixed it in timelib. It will sync into PHP shortly: https://github.com/derickr/timelib/commit/483fbab1c83981a5573a1123f49ce5033da8775f

Derick supposes it will hit PHP versions 5.5.31 / 5.6.15 / 7.0.0RC6. Since we still support 5.4 (and versions earlier than 5.5.31 and 5.6.15) in this library, we may not be able to use this functionality for a while. I'm investigating other options.

thomasmuehlschlegel:
Another try:

```
$unixTime = ($this->getTimestamp() - 0x01b21dd213814000) / 1e7;
$microTime = abs($unixTime * 1e6 - (int)$unixTime * 1e6);

return new \DateTime(date('Y-m-d H:i:s', (int)$unixTime) . '.' . (int)$microTime);
```

ramsey:
I'm still working on this, but it's causing test failures, and I want to be careful, since changing the tests without considering the consequences in these cases could break software that depends on this library.

See #93 for the changes.

ramsey:
Please see version 4.0.0-alpha1. The `getDateTime()` method now returns a `\DateTimeInterface` instance that includes the microseconds.

azerpas:
> Please see version 4.0.0-alpha1. The `getDateTime()` method now returns a `\DateTimeInterface` instance that includes the microseconds.

How can we get 4.0.0 through composer? Waiting for that release for a project. 👍

ramsey:
While the 4.0.0-alpha1 API is fairly stable, I don't recommend using it in production until the 4.0.0 release. That said, if you'd like to test it, I would recommend using this in your `composer.json`:

    "ramsey/uuid": "4.0.0-alpha1 as 4.0.0"

PR #93, opened 2015-10-25
Add microtime to the returned DateTime object

ramsey:
Discussed in #90. Currently failing tests and want to be careful not to change tests for the sake of making this work, without considering the implications of the changes to software that depends on this library.

### DIFF
--- a/src/DegradedUuid.php
+++ b/src/DegradedUuid.php
@@ -33,12 +33,21 @@ class DegradedUuid extends Uuid
         $time = $this->converter->fromHex($this->getTimestampHex());
 
         $ts = new \Moontoast\Math\BigNumber($time, 20);
-        $ts->subtract('122192928000000000');
-        $ts->divide('10000000.0');
-        $ts->round();
+        $ts->subtract('122192928000000000'); // 0x01b21dd213814000
+        $ts->divide('10000000.0'); // 1e7
         $unixTime = $ts->getValue();
 
-        return new \DateTime("@{$unixTime}");
+        $microTime = new \Moontoast\Math\BigNumber($unixTime, 20);
+        $microTime->multiply('1000000.0'); // 1e6
+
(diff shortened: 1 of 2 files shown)


mcneely:
Perhaps a config setting to toggle microtime on or off? You can keep the original test, toggle it on, then add a test to check it toggled on then.

ramsey:
@mcneely I'm considering setting the minimum required PHP version for the 4.x series to 5.6.15, which would mean all supported PHP versions properly support microtime, and I don't have to worry about supporting it for older PHP versions.

azerpas:
Any plan to add it soon?

kalifg:
I was just looking at this and there was only one failing test:
```
There was 1 failure:

1) Ramsey\Uuid\Test\UuidTest::test32BitMatch64BitForOneHourPeriod
Failed asserting that 1355184000 matches expected 1355184001.0.

/Users/michaeldwyer/src/uuid/tests/src/UuidTest.php:1281
```

This only occurs when `$usec` is >= 500000 and is simply because `round` rounds up when the fractional portion of the timestamp is >= .5.  However `DateTime::getTimestamp()` simply drops the fractional portion.  Changing this

```php
$testTime = round($currentTime + ($usec / 1000000));
```
to this
```php
$testTime = floor($currentTime + ($usec / 1000000));
```

I looked into master, and this isn't a problem because in `Uuid::getDateTime()`, `Uuid::getTimestamp()` returns `1355184000.5`, which `number_format` turns into `1355184001`, matching the logic of `round`.

You said you are reluctant to change the test, but in this case I think you might consider it a correction since the current behavior of returning a rounded-up `DateTime` when presented with microseconds is inconsistent with `DateTime`'s choice to drop them.

ramsey:
This is coming in 4.0.0. I’ve already addressed the rounding issue you mentioned in the `4.x` branch, and I’ll be adding the microtime support soon. Thanks for following up!

PR #94, opened 2015-12-17
Added support for symfony/console v3

pkruithof:
See puli/discovery#19 for discussion

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,5 @@
+sudo: false
+
 language: php
 
 php:
@@ -7,9 +9,15 @@ php:
   - 5.6
   - hhvm
 
+matrix:
+  include:
+    - php: 5.6
+      env: COMPOSER_FLAGS='--prefer-lowest --prefer-stable'
+  fast_finish: true
+
 before_script:
   - composer self-update
-  - composer install --dev --prefer-source
+  - composer update $COMPOSER_FLAGS --prefer-source
 
 script: ./vendor/bin/phpunit --coverage-text --coverage-clover ./build/logs/clover.xml
 
--- a/composer.json
+++ b/composer.json
@@ -24,9 +24,9 @@
     },
     "require-dev": {
         "moontoast/math": "~1.1",
-        "symfony/console": "~2.3",
+        "symfony/console": "~2.3|~3.0",
         "doctrine/dbal": ">=2.3",
-        "phpunit/phpunit": "~4.1",
+        "phpunit/phpunit": "~4.1|~5.0",
         "satooshi/php-coveralls": "~0.6",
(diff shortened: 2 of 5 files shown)


.travis.yml
```diff
@@ -7,9 +7,15 @@ php:
...
+matrix:
+  include:
+    - php: 5.6
+      env: COMPOSER_FLAGS='--prefer-lowest --prefer-stable'
+  fast_finish: true
+
 before_script:
   - composer self-update
-  - composer install --dev --prefer-source
+  - composer update $COMPOSER_FLAGS --prefer-source -n
```
GrahamCampbell: `-n` is not needed

GrahamCampbell: travis already force enable it using environment variables

composer.json
```diff
@@ -24,7 +24,7 @@
     },
     "require-dev": {
         "moontoast/math": "~1.1",
-        "symfony/console": "~2.3",
+        "symfony/console": "~2.3|^3.0",
```
GrahamCampbell: should be `~` not `^`

composer.json
```diff
@@ -24,7 +24,7 @@
...
         "doctrine/dbal": ">=2.3",
         "phpunit/phpunit": "~4.1",
```
GrahamCampbell: would make sense to allow `~5.0` here too

.travis.yml
pkruithof: But it's non-interactive, composer could ask a question or something (maybe about GitHub tokens), if not now then maybe in a future release?

pkruithof: Ah, nevermind, I did not know that. Are they using an environment variable or something?

GrahamCampbell: > But it's non-interactive, composer could ask a question or something (maybe about GitHub tokens), if not now then maybe in a future release?

No, travis already force it into non-interactive mode.

ramsey: I don't understand the use of "update" instead of "install."

ramsey:
The tests are [passing here](https://travis-ci.org/ramsey/uuid/builds/97471220) because they're grabbing version 2.8.0 of symfony/console (and one is grabbing 2.3.0 for `--prefer-lowest`). The problem I mentioned in puli/discovery#19 is that the tests fail with errors when symfony/console 3.0 is used.

pkruithof:
This build uses 3.0 and passes: https://travis-ci.org/ramsey/uuid/jobs/97472351

The failing build is because `--prefer-lowest` is not an option for `install`, but for `update`, so that's apparently the reason for `update` instead of `install`.

ramsey:
All looks good. Thanks for your help!

PR #99, opened 2016-02-06
Migrated isValid to a ValidatorInterface

bradynpoulsen:
Almost all of the static methods on the `Uuid` class references the `UuidFactory`. I love this approach and believe it should continue with `isValid(string $uuid) : boolean` as well.

In my projects, I pass around an instance of UuidFactory for dependency injection and type-hint against UuidFactoryInterface in case there were ever a time I wanted to use another implementation. Everything I needed was available in the factory EXCEPT the `Uuid::isValid` method. This won't break any users that utilize the Uuid method or `VALID_PATTERN` constant as they are still there, but instead reference UuidFactoryInterface.


bradynpoulsen:
**NOTE** this will be a breaking change for anyone with classes that implement `UuidFactoryInterface` but do not have the `isValid($uuid)` method

src/UuidFactory.php
```diff
@@ -178,6 +183,29 @@ public function fromInteger($integer)
...
+     *
+     * @param  string  $uuid The string to check
+     * @return boolean       Returns true if the string represents a valid UUID; otherwise, returns false
+     */
+    public function isValid($uuid)
+    {
+        // assert or InvalidArgumentException should be used for the integrity of the argument
+        // These statements give helpful indication during debugging
+        assert(func_num_args() === 1);
+        assert('is_string($uuid)');
```
ramsey: Please remove these `assert()` statements. I don't think they provide any value beyond the `if` statement below them. The `if` statement still verifies that at least one argument was passed in and that it matches a value UUID pattern, which must be a string.

ramsey:
Thanks for this pull request. I'm going to give it some more thought before merging it in, since it is a breaking change.

I need to consider whether this can go immediately into a 3.2.0 release or whether I need to plan it for 3.3.0, while doing an intermediate 3.2.0 release that somehow warns users that `isValid()` will be required on implementations of UuidFactoryInterface in the next version. I'm not sure the best way to do that, and I'm not sure it's necessary, since it's difficult to tell how many implementations this will actually affect, if any.

src/UuidFactory.php
```diff
@@ -178,6 +183,29 @@ public function fromInteger($integer)
...
+
+        $uuid = str_replace(array('urn:', 'uuid:', '{', '}'), '', $uuid);
+
+        // This should probably be moved to the UuidFactory from Uuid
+        if ($uuid === Uuid::NIL || preg_match('/' . self::VALID_PATTERN . '/', $uuid)) {
```
bradynpoulsen: @ramsey What are your thoughts on `Uuid::NIL`? Should it be referenced as a constant of the `UuidFactory` as well?

ramsey: I prefer to keep the `NIL` constant on the Uuid class. The Uuid class is the base class of the library. All the other namespace constants, etc. live there. Most use-cases involve simply invoking the static methods on the Uuid class to get UUIDs. The factory and interface are provided to allow others a convenient way to replace certain generators with their own implementations, but not many are using these in practice.

bradynpoulsen:
That's one of the harder decisions with semantic versioning is adding a new method to an interface like this. While it won't break existing method implementations, it may break the overall class implementation for the lack thereof.

aztech-dev:
Moving isValid() to the factory classes bothers me a bit. UUID validation is unrelated to the creation process, and it introduces a design issue. All factories will now be required to implement an almost-invariant algorithm, leading to either unnecessary code duplication, or worse, an AbstractFactory that all implementers will extend to avoid implementing the validation code.

It's probably smoother to push your refactoring further by extracting a UuidValidator class (not sure an interface is useful here) from that code, and adding a `getValidator()` method on `UuidFactoryInterface`. This has the added benefit of solving your use case, and letting UuidFactoryInterface implementers easily return the default validator or a custom one as needed.

However, that does not solve the BC issue :(

ramsey:
I like @aztech-dev's suggestion of the UuidValidator class with a `getValidator()` method on UuidFactoryInterface. @bradynpoulsen, thoughts?

For now, I'm adding this to the 4.0.0 milestone.

bradynpoulsen:
@ramsey I agree. I just wanted to prioritize moving away from static-only availability :) @aztech-dev good catch

src/FeatureSet.php
```diff
@@ -14,30 +14,31 @@
 
 namespace Ramsey\Uuid;
 
-use Ramsey\Uuid\Converter\TimeConverterInterface;
-use Ramsey\Uuid\Generator\PeclUuidTimeGenerator;
-use Ramsey\Uuid\Provider\Node\FallbackNodeProvider;
-use Ramsey\Uuid\Provider\Node\RandomNodeProvider;
-use Ramsey\Uuid\Provider\Node\SystemNodeProvider;
+use Ramsey\Uuid\Builder\DefaultUuidBuilder;
```
bradynpoulsen: Sorry, PHPCompanion for Sublime Text must've automatically sorted the use statements when it imported the `ValidatorInterface`

src/FeatureSet.php
```diff
@@ -96,6 +97,11 @@ class FeatureSet
     private $timeGenerator;
 
     /**
+     * @var ValidatorInterface
+     */
+    private $validator;
```
bradynpoulsen: I already removed this from the `UuidFactory` since it had nothing to do with creating UUIDs. Should I remove it from here as well since it's dead use unless some interfaces with `FeatureSet` directly to get/set the validator?

src/Validator/Validator.php
```diff
@@ -0,0 +1,44 @@
...
+     * @param string $uuid
+     * @return bool Returns TRUE if the string was validated as a valid UUID or FALSE on failure
+     */
+    public function validate($uuid)
+    {
+        assert(func_num_args() === 1);
+
+        $uuid = str_replace(array('urn:', 'uuid:', '{', '}'), '', $uuid);
+
+        // This should probably be moved to the UuidFactory from Uuid
```
bradynpoulsen: TODO: remove this suggestion comment

src/FeatureSet.php
ramsey: Our standard pattern has been to move these things onto `UuidFactoryInterface`. Then, when needing to change the dependencies, we get a factory, set the values there, and then when the methods on `Uuid` are called, they take advantage of that. We should follow the same pattern here.

For example:

``` php
$factory = Uuid::getFactory();
$factory->setValidator(new MyValidator());

if (Uuid::isValid($foo)) {
    // do something
}
```

At this point, `Uuid` is now using your own validator. The internals of `isValid()` would look something like this:

``` php
public static function isValid($uuid)
{
    return self::getFactory()->getValidator()->validate($uuid);
}
```

I think this is what @aztech-dev was suggesting on March 22.

src/Validator/Validator.php
```diff
@@ -0,0 +1,44 @@
...
+ */
+class Validator implements ValidatorInterface
+{
+    /**
+     * Validate that a string represents a UUID
+     *
+     * @param string $uuid
+     * @return bool Returns TRUE if the string was validated as a valid UUID or FALSE on failure
+     */
+    public function validate($uuid)
```
ramsey: Should this accept only a string, or could `$uuid` be either a string or a `UuidInterface` object?

src/FeatureSet.php
ramsey: No worries, but typically I request coding style and whitespace changes to be in separate PRs, to avoid noise. :-)

src/FeatureSet.php
bradynpoulsen: Ok. I will then have to add `getValidator()`/`setValidator()` as part of the factory interface, correct? That's what I had and then the description of the factory said it was common functionality to create UUIDs, not necessarily interpret/validate them

src/Validator/Validator.php
bradynpoulsen: @ramsey Are you able to fulfill the UuidInterface and still have an invalid UUID?

Issue #80, opened 2015-08-27
Uuid:uuid4() collisions


giorgiosironi:
We are generating about 1M UUID4 a day, and we are getting several hundred collisions a day, such as:
```
[2015-08-26 21:29:19 +0200] [production-onebipws-1.apache - 17819] [DEBUG] [Request] 
time_total=0.145
"request_id":"fdfb98c1-4367-4f22-b68a-d7cdaedcc069" 

[2015-08-27 00:36:16 +0200] [production-onebipws-1.apache - 17819] [DEBUG] [Request] 
time_total=0.016
"request_id":"fdfb98c1-4367-4f22-b68a-d7cdaedcc069"
```
The issue seem to be correlated with the same Apache process regenerating the same UUID after several hours. It also seem to be correlated with particular EC2 machines which presents the problem.

We checked to have `openssl_random_pseudo_bytes` and if it was using a strong algorithm:
```
root@dev-all-onebip:~/projects/.../onebip-ultimate$  (master) # php -r "var_dump(function_exists('openssl_random_pseudo_bytes'));"                                      
bool(true)
root@dev-all-onebip:~/projects/.../onebip-ultimate$  (master) # php -r 'openssl_random_pseudo_bytes(16, $strong); var_dump($strong);'
bool(true)
```

How can we debug this problem?

ircmaxell:
Which version of UUID are you using? Can you try switching to master?

giorgiosironi:
We are on `2.8.1`, but from what I see the code for uuid4() is identical to master:
```
 public static function uuid4()
{
$bytes = self::generateBytes(16);
// When converting the bytes to hex, it turns into a 32-character
// hexadecimal string that looks a lot like an MD5 hash, so at this
// point, we can just pass it to uuidFromHashedName.
$hex = bin2hex($bytes);
return self::uuidFromHashedName($hex, 4);
}

 private static function generateBytes($length)
{
if (self::hasOpensslRandomPseudoBytes()) {
return openssl_random_pseudo_bytes($length);
}
```
Here is a sample of the duplicated UUIDs:
https://gist.github.com/giorgiosironi/f1ce4682868ca6a6279d

ramsey:
I wonder if you'll experience the same if you switch to using @ircmaxell's RandomLib.

Using `master` or `3.0.0-alpha3`, you can do so like this:

``` php
$uuidFactory = new \Ramsey\Uuid\UuidFactory();
[...]

PR #100, opened 2016-03-15
Drop OpenSSL support and use paragonie/random_compat

ramsey:
Fixes issue #80 for the 2.x series

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
     },
     "require": {
         "php": ">=5.3.3",
-        "paragonie/random_compat": "dev-without-openssl"
+        "paragonie/random_compat": "dev-master"
     },
     "require-dev": {
         "moontoast/math": "~1.1",
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,8 @@
         "source": "https://github.com/ramsey/uuid"
     },
     "require": {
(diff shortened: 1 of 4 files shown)


composer.json
```diff
@@ -20,7 +20,8 @@
         "source": "https://github.com/ramsey/uuid"
     },
     "require": {
-        "php": ">=5.3.3"
+        "php": ">=5.3.3",
+        "paragonie/random_compat": "^2.0"
```
GrahamCampbell: Best to allow `^1.0|^2.0` otherwise will will conflict with just about everything atm.

GrahamCampbell: For example, you won't be able to install this on laravel or symfony.

ramsey: Good point. I'll do that.

ramsey: Only problem is that it will continue to cause UUID collision problems for folks who are using something that's not updated to 2.0, so I'm a little hesitant to do this, but I'll do it nonetheless, so that I don't cause breaks.

ramsey: Actually, it shouldn't really be much of an issue, since most systems will match a CSPRNG in paragonie/random_compat before it gets to the OpenSSL block, so not a big deal.

PR #101, opened 2016-03-15
Drop the use of OpenSSL as a fallback and use paragonie/random_compat

ramsey:
Fixes issue #80 for the 3.x series


composer.json
```diff
@@ -25,7 +25,8 @@
         "source": "https://github.com/ramsey/uuid"
     },
     "require": {
-        "php": ">=5.4"
+        "php": ">=5.4",
+        "paragonie/random_compat": "^2.0"
```
GrahamCampbell: Best to allow `^1.0|^2.0` otherwise will will conflict with just about everything atm.

GrahamCampbell: For example, you won't be able to install this on laravel or symfony.

Issue #86, opened 2015-10-03
Collisions occurring in COMB sequential UUID test


ramsey:
The presence of Xdebug appears to slow things down significantly enough that COMB sequential UUIDs generated during tests have no collisions, but if tests run in an environment without Xdebug, then tests will almost always produce duplicate COMB UUIDs.

To reproduce (using a VirtualBox VM with Vagrant):

    git clone https://github.com/ramsey/uuid.git
    cd uuid/
    vagrant init ubuntu/trusty64
    vagrant up
    vagrant ssh
    cd /vagrant
    sudo apt-get install php5 php5-curl
    curl -sS https://getcomposer.org/installer | php
    ./composer.phar install
    ./vendor/bin/phpunit --verbose

You will see a consistent test failure like this:

    1) Ramsey\Uuid\UuidTest::testUuid4Comb
    Failed asserting that '30303030-3030-4030-b030-8351f882227f' is greater than '30303030-3030-4030-b030-8351f8822280'.

    /vagrant/tests/UuidTest.php:839

Now, install Xdebug and run the test again:

    sudo apt-get install php5-xdebug
    ./vendor/bin/phpunit --verbose

Tests will consistently pass.

Remove Xdebug, and tests consistently fail again:

    sudo apt-get remove php5-xdebug
    ./vendor/bin/phpunit --verbose

We need to revisit COMB generation to see what can be done to ensure consistent, sequential UUIDs.

aztech-dev:
Interesting. I'm not getting that behavior at all, with or without XDebug. May be related to the performance of the machine on which the tests are running:

```
thibaud@thibaud:uuid$ sudo php5enmod xdebug
thibaud@thibaud:uuid$ phpunit
PHPUnit 4.6.6 by Sebastian Bergmann and contributors.

Configuration read from uuid/phpunit.xml.dist

........S......................................................  63 / 109 ( 57%)
......................................S.......

Time: 3.94 minutes, Memory: 13.00Mb

OK, but incomplete, skipped, or risky tests!
Tests: 109, Assertions: 67397, Skipped: 2.

Generating code coverage report in Clover XML format ... done

Generating code coverage report in HTML format ... done
thibaud@thibaud:uuid$ sudo php5dismod xdebug
thibaud@thibaud:uuid$ phpunit
PHPUnit 4.6.6 by Sebastian Bergmann and contributors.

Configuration read from uuid/phpunit.xml.dist

[...]

PR #102, opened 2016-03-16
Implemented timestamp first and timestamp last comb generators

ghola:
I know @aztech-dev is already working on this as per issue #86 and it's possible that his implementation will be more complete. The goal of this patch is to provide a quick implementation for combs usable with MySQL. I did not take little endian systems into account.

BC may be affected too, because I refactored the comb generator implementation using the template pattern and replaced the conditional with polymorphism (passing a type into the CombGenerator constructor would have lead to conditional code and breaking of the open/closed principle with each new comb variation).


aztech-dev:
Well actually :D I haven't made much progress on that P/R for two major reasons:

- Time available
- I have doubts as to whether a generator is the best place to do handle that aspect

All in all, a CombGenerator is only a random generator that replaces some of the random bytes by a timestamp. My gut feeling is pretty much now that the logic of inverting the position of the random bytes and the timestamp should rather be placed in encoder/decoders (pretty much in the same way that GUIDs are currently handled). 

I meant to open a discussion about this, but haven't gotten around to it, so your P/R is pretty much forcing me into doing it :) Thoughts ?

ghola:
From what I see encoders are meant to be used with any generator. Would an encoder that inverts the position of byte sequences be compatible with all uuid versions? If not, then I feel that we make it harder on the end user because now he needs to know which encoder is compatible with which generator.

ramsey:
I restarted the HHVM build. All other builds passed. That one just seemed to time out.

First, thanks so much for your contribution and interest in this library! :-)

I worry about making **CombGenerator** an abstract class, since that definitely does break BC, though it's likely it won't affect a lot of folks. Additionally, we've been following the [Symfony naming conventions](http://symfony.com/doc/current/contributing/code/standards.html#naming-conventions) for the most part, so the name should become **AbstractCombGenerator**.

Maybe a good compromise is to introduce an **AbstractCombGenerator** and keep **CombGenerator** but have it extend the abstract class? This avoid the huge BC break caused by changing the class to abstract.

As for @aztech-dev's concerns, I'll let y'all hash that out here. I could go either way on the generator vs. encoder argument right now.

aztech-dev:
@ghola Yes it would work with any generator. I agree with the "makes it harder on the user" point, so I guess the generator solution is the better one when considering DX.

Regarding the abstract class, I'll side with @ramsey (in spite of my public hatred for Abstract prefixes and Interface suffixes) for the sake of BC :)

aztech-dev:
Edit: I misread your objection. Yes, an encoder is meant to be able to handle any UUID, regardless of its version. For the most part, they're just re-arranging some bytes and converting to string, so "incompabilities" can occur in the sense that applying a given encoder would not really make any sense depending on the UUID type (ie. using a CombEncoder on a V1 UUID), but it clearly will not (and **must not**) fail.

ghola:
@aztech-dev Yes, the V1 UUID was what I had in mind too. I would argue that the main concern with an encoder that inverts byte positions lies not in the possibility of failure but in that it would produce an incorrect output. The developer relies on the fact that he can extract the timestamp from a V1 at any time, so he might be taken by surprise when he notices that the codec he uses messes some UUIDs up.

@ramsey I made another push following your suggestions.

ghola:
Removed timestampBytes as a member and made it a constant. Also removed redundant code.
Made this in order to be more consistent with the move of the member initialization out of the constructor and with the position taken in PR #43.

aztech-dev:
That should not be an issue as the `CodecInterface` contract is that whatever is returned by the `encode` method must be decodable by the `decode` method. So it would only be an issue if you encode with one codec type and decode with another one (ie encode with standard string code, and try to decode with the GUID codec - it will obviously yield incorrect results, and the responsibility of encoding/decoding with the same codec falls outside of the scope of this library IMO).

By default, the library always uses the same codec, and using another codec is the responsibility of the user.

What bothers me in the generator approach is that it ties your UUID's internal representation to your storage implementation. The advantage of using codecs is that your internal UUID representation never varies, and I can convert it to whatever target database I want. If I generate a COMB optimized for MySQL, and for some reason, end up having to store that same COMB in MSSQL, I will loose all performance gains since it will not be optimized for the second engine, unless I write a Codec to convert the UUID to the proper representation for the target engine.

So in the end, if you end up needing to write codecs for each target database, you might as well use codecs directly instead of relying on the generator.

aztech-dev:
Also to get back on your statement that you did not account for endianness, it does not matter, since it's only significant for GUID's (that have two fields that are host-endianness dependant), and that is handled via the GUID codec.

ramsey:
> What bothers me in the generator approach is that it ties your UUID's internal representation to your storage implementation.

I've been giving this a lot of thought, and I'm inclined to side with @aztech-dev on this. Changing to use a codec rather than a generator does present a few problems, though.

With the current structure using a generator, this is the pattern of code users may use to select the generator they wish to use:

``` php
$uuidFactory = new \Ramsey\Uuid\UuidFactory();
$uuidFactory->setRandomGenerator(new \Ramsey\Uuid\Generator\TimestampFirstCombGenerator());
\Ramsey\Uuid\Uuid::setFactory($uuidFactory);

$uuid = \Ramsey\Uuid\Uuid::uuid4();
```

If we turn it into a codec, it gets more complicated:

``` php
class MyCombFeatureSet extends \Ramsey\Uuid\FeatureSet
{
    protected function buildCodec($useGuids = false)
    {
        return new \Ramsey\Uuid\Codec\TimestampFirstCombCodec();
    }
}

$uuidFactory = new \Ramsey\Uuid\UuidFactory(new MyCombFeatureSet());
\Ramsey\Uuid\Uuid::setFactory($uuidFactory);

$uuid = \Ramsey\Uuid\Uuid::uuid4();
```

So, if we decide to turn this into a codec, I recommend adding a `setCodec()` method to `Ramsey\Uuid\UuidFactory`, which will simplify its use to:

``` php
$uuidFactory = new \Ramsey\Uuid\UuidFactory();
$uuidFactory->setCodec(new \Ramsey\Uuid\Codec\TimestampFirstCombCodec());
\Ramsey\Uuid\Uuid::setFactory($uuidFactory);

$uuid = \Ramsey\Uuid\Uuid::uuid4();
```

I think this is more fitting with how COMB should be used, given that a COMB is a combined GUID/timestamp, and we already have support for GUIDs through the use of a codec, rather than a generator.

Adding a `setCodec()` method to the factory also gives users the ability to do this:

``` php
$uuidFactory = new \Ramsey\Uuid\UuidFactory();
$uuidFactory->setCodec(new \Ramsey\Uuid\Codec\GuidStringCodec());
\Ramsey\Uuid\Uuid::setFactory($uuidFactory);

$uuid = \Ramsey\Uuid\Uuid::uuid4();
```

Instead of this (which doesn't clearly show your intent in the code, IMO):

``` php
$featureSet = new \Ramsey\Uuid\FeatureSet(true);
$uuidFactory = new \Ramsey\Uuid\UuidFactory($featureSet);
\Ramsey\Uuid\Uuid::setFactory($uuidFactory);

$uuid = \Ramsey\Uuid\Uuid::uuid4();
```

A final problem with these changes is that of semantic versioning. If we introduce a COMB codec and remove the COMB generator, that is an incompatible API change, which will necessitate a bump to 4.0. However, I don't think it's time to make this leap. So, my proposal would be to move all the functionality into a codec, while leaving `Ramsey\Uuid\Generator\CombGenerator` behind as a wrapper around the codec functionality, and we mark it with a deprecation annotation, with the class to be removed entirely in 4.0.

Thoughts?

aztech-dev:
> Instead of this (which doesn't clearly show your intent in the code, IMO):

Couldn't agree more on the intent issue.

>  So, my proposal would be to move all the functionality into a codec, while leaving Ramsey\Uuid\Generator\CombGenerator behind as a wrapper around the codec functionality, and we mark it with a deprecation annotation, with the class to be removed entirely in 4.0

Not sure I understand how this approach would be implemented. Another solution (one that I can wrap my head around -- may be the same as your suggestion worded differently) would be to deprecate the current CombGenerator and introduce a new invariant one that relies on codecs. But I agree that a major version bump is not desirable now.

ramsey:
> Another solution (one that I can wrap my head around -- may be the same as your suggestion worded differently) would be to deprecate the current CombGenerator and introduce a new invariant one that relies on codecs.

I think we're saying the same thing. :-)

ramsey:
@ghola, sorry to pile all this on you, but how does all that sound?

ghola:
@aztech-dev convinced me with the difference between internal and external representation, so maybe the codecs are the better way to do this.

But I have a concern related to codec chainability. If I want to change the endianness and the timestamp position, ideally I would combine two existing codecs instead of making a new one.

Also, do we really need to deprecate the current CombGenerator? We could assume that the internal representation of the comb is the one already implemented (with timestamp at the end) and then we would have two new codecs: one would reverse the bytes position and the other would be a noop.

ghola:
New push with two codecs, their tests and BC for CombGenerator. Let me know if this is what you had in mind.

src/Codec/TimestampLastCombCodec.php
```diff
@@ -0,0 +1,23 @@
...
+ * @link https://packagist.org/packages/ramsey/uuid Packagist
+ * @link https://github.com/ramsey/uuid GitHub
+ */
+namespace Ramsey\Uuid\Codec;
+
+/**
+ * TimestampLastCombCodec encodes and decodes COMB UUIDs which have the timestamp as the last 48 bits.
+ * To be used with MSSQL.
+ */
+class TimestampLastCombCodec extends StringCodec
```
aztech-dev: Is it really necessary to add this extra class ?

src/Codec/StringCodec.php
```diff
@@ -29,7 +29,7 @@ class StringCodec implements CodecInterface
     /**
      * @var UuidBuilderInterface
      */
-    private $builder;
+    protected $builder;
```
aztech-dev: It's a bit of nitpicking, but I find it better to leave the property private and add a protected getter to prevent accidental overwrites.

aztech-dev:
Looks great ! Just wondering whether it's really opportune to have the extra timestamp last encoder, considering we can achieve the same effect using the string codec directly.

aztech-dev:
> But I have a concern related to codec chainability. If I want to change the endianness and the timestamp position, ideally I would combine two existing codecs instead of making a new one.

Currently, codec's are not designed with that in mind, so it's doable, but it will require extra encode/decode operations and won't be very efficient as a result.

ghola:
I have no problem removing the TimestampLastCombCodec. My only argument for having it was that it would be more obvious for the user. I imagine a MSSQL user that looks at the TimestampFirst one and wonders where the TimestampLast is. To answer that question he would either come and post an issue or would need to go through the source code of the CombGenerator and see for himself how that gets generated. Wasted time in both occasions.

Let me know if you want it removed.

src/Codec/StringCodec.php
ghola: Just pushed a new one with your request.

aztech-dev:
> My only argument for having it was that it would be more obvious for the user

Fair enough :)

ramsey:
@ghola, can you take a look and fix the conflicts here, please? I released 3.3.0, and it looks like some of those changes caused conflicts with your pull request. Thanks!

ghola:
@ramsey Done.

ramsey:
I'm not sure why there are more conflicts, but can you fix these, as well, @ghola? I'll review and accept this PR next, so I'm not going to pull anything else in that will cause more conflicts. Sorry for the inconvenience. :-)

ramsey:
@ghola, ping. I'd like to get this merged in. Thanks!

ghola:
@ramsey Done.

ramsey:
Thanks a bunch!

PR #103, opened 2016-03-17
Increase test coverage

jmauerhan:
I've added some tests for specific classes and methods that were previously not covered. The tests are all isolated and single responsibility - so each test is either for collaboration with a dependency, correctness, or contract of public API. 

The following classes have increased coverage:
- CombGenerator
- DefaultUuidBuilder
- DegradedUuidBuilder
- PhpTimeConverter
- MtRandGenerator
- RandomLibAdapter


ramsey:
Thought I'd point out that I'm not using the Travis CI version of PHPUnit but am instead including PHPUnit [as part of require-dev](https://github.com/ramsey/uuid/blob/e4e38b405963e6b49ac1ea10b5e22cbb76158c9e/composer.json#L33) and [running it from the vendor directory](https://github.com/ramsey/uuid/blob/47964f2cca3876454904c807831ebeb5650fb264/util/run-tests.sh#L36) as part of my test runner.

The `expectException()` method was added in PHPUnit 5.2. Since this library supports PHP 5.4+, the tests that ran on PHP 5.4 and 5.5 used PHPUnit 4.8.24 and the tests that ran on 5.6, 7.0, and HHVM used PHPUnit 5.2.12.

It's frustrating, but that's why your `expectException()` calls didn't work.

Going to review the rest, but so far everything looks great. Thanks!

ramsey:
Also, the PHPUnit 5.x series requires PHP 5.6+, which is why I need to specify it in `composer.json` as `^4.7|^5.0`, so that it can resolve the dependencies properly when installing on PHP 5.4 and 5.5. :-)

ramsey:
All looks great! Thanks!

jmauerhan:
@ramsey Thanks for explaining the issue I had with travis - I always run phpunit from the vendor (work on tons of projects with different versions) and the version I got when I installed uuid was `5.2.10` - but I'm running 7 locally. I realize now it was just the 5.5 and 5.4 builds that would have failed, but I guess I saw those first and didn't look at the others, so I didn't make the connection :)

PR #104, opened 2016-03-23
Add src/Generator test coverage

jmauerhan:
- Added tests for the rest of the classes in the Generator namespace. 
- Added AspectMock to create test doubles of php methods, uses a namespacing trick to create doubles. This only works on methods that aren't specified as in global namespace (ie: `\random_bytes()` instead of `random_bytes()` - fortunately this worked on all of these. 
- Added some checks for HHVM because AspectMock does not work on HHVM. Talked to @ramsey who said we could exclude those tests on HHVM. 
- Increased coverage to 96% according to coveralls


ramsey:
This is all great. Thanks!

I'm going to start using AspectMock myself, after seeing it here. Awesome stuff!

PR #108, opened 2016-03-24
Test coverage providers

jmauerhan:
- Added tests for all Providers. All tests are isolated correctness/contract/collaboration tests. 
- Moved the AspectMock teardown into the TestCase teardown. This won't affect anything that isn't using AspectMock, and will prevent forgetting to clean AspectMock when a test is over.

### DIFF
--- a/tests/src/Provider/Node/SystemNodeProviderTest.php
+++ b/tests/src/Provider/Node/SystemNodeProviderTest.php
@@ -19,8 +19,8 @@ class SystemNodeProviderTest extends TestCase
 
         $this->assertTrue(ctype_xdigit($node), 'Node is not a hexadecimal string. Actual node: ' . $node);
         $this->assertTrue(strlen($node) === 12, 'Node is 12 characters long. Actual length: ' . strlen(
-                $node
-            ) . PHP_EOL . ' Actual node: ' . $node);
+            $node
+        ) . PHP_EOL . ' Actual node: ' . $node);
     }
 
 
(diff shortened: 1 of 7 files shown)


jmauerhan:
As a note, some of these things were covered by other tests, but those tests are not isolated - which means there were not any specific tests for these item's contracts and collaboration with their dependencies. I find that these specific isolated tests provide better security when refactoring and make it easier to make non-breaking changes to both the consumers and providers for the system they test. So this PR does not add a ton of new coverage, but was intended to add improved quality of the coverage. 

Thanks :)

tests/src/Provider/Node/SystemNodeProviderTest.php
```diff
@@ -0,0 +1,136 @@
...
+     * @runInSeparateProcess
+     * @preserveGlobalState disabled
+     */
+    public function testGetNodeReturnsSystemNodeFromMacAddress()
+    {
+        $provider = new SystemNodeProvider();
+        $node = $provider->getNode();
+
+        $this->assertTrue(ctype_xdigit($node), 'Node is not a hexadecimal string. Actual node: ' . $node);
+        $this->assertTrue(strlen($node) === 12, 'Node is 12 characters long. Actual length: ' . strlen(
```
ramsey: Should the message be "Node is *not* 12 characters long." ?

aztech-dev: or "Node *should* be 12 characters long, got %d instead"

ramsey:
@jmauerhan, I've reviewed the pull request, and all looks great. I think the only change is that test message mentioned above. Once you make that change, I'll merge it in. Thanks!

tests/src/Provider/Node/SystemNodeProviderTest.php
jmauerhan: Good catch guys thanks 😊 I'll make that asap

PR #118, opened 2016-04-27
Codec to store UUID in an optimized way for InnoDB

### LINKED ISSUE #117: Store UUID in an optimized way for InnoDB
From Karthik Appigatla on the Percona blog ("[Store UUID in an optimized way](https://www.percona.com/blog/2014/12/19/store-uuid-optimized-way/)"):

> * UUID has 36 characters which makes it bulky.
> * InnoDB stores data in the PRIMARY KEY order and all the secondary keys also contain PRIMARY KEY. So having UUID as PRIMARY KEY makes the index bigger which can not be fit into the memory
> * Inserts are random and the data is scattered.

He goes on to "explain how to store UUID in an efficient way by re-arranging timestamp part of UUID."

While, ramsey/uuid provides COMB sequential UUIDs, these do not rearrange the timestamp part of the UUID to optimize for storage in InnoDB. I'd like to consider the addition of a codec that would help optimize UUIDs as primary keys for the InnoDB engine. Perhaps other database engines require similar optimization techniques, so we should thing about those, as well.

The problem with rearranging the timestamp part of a UUID is that the version number is lost, since it is shuffled around, and there would be no way to determine the original state of the UUID. Whether or not this is important is up for debate.

barryvdh:
Fixes #117: Store UUID in an optimized way for InnoDB

Adds a Coded that extends the StringCodec. The string encoding is unchanged, but the binary output is rearranged so that it increases over time (instead of random). The UUID fields itself have not change, only the ordering. No data is lost, so decoding back will give the original string.

Example use case would be to store the data in an optimized way, but show the original UUID in the admin interface. See below for examples, if you run it, you'll see that you can convert the UUID back and forth + the binary values are increasing, as suggested by https://www.percona.com/blog/2014/12/19/store-uuid-optimized-way/

The tests should cover this also.

Example usage:

```php
use Ramsey\Uuid\Uuid;
use Ramsey\Uuid\Codec\OrderedTimeCodec;
use Ramsey\Uuid\Exception\UnsatisfiedDependencyException;

// Generate a version 1 (time-based) UUID object
$uuid1 = Uuid::uuid1();
echo $uuid1->toString() . "\n"; // i.e. 58e0a7d7-eebc-11d8-9669-0800200c9a66
echo bin2hex($uuid1->getBytes()) . "\n"; // i.e. 58e0a7d7eebc11d896690800200c9a66

// Generate an optimized version 1 (time-based) UUID object
$factory = new \Ramsey\Uuid\UuidFactory();
$codec = new OrderedTimeCodec($factory->getUuidBuilder());
$factory->setCodec($codec);
Uuid::setFactory($factory);

$uuid1 = Uuid::uuid1();
echo $uuid1->toString() . "\n"; // i.e. 58e0a7d7-eebc-11d8-9669-0800200c9a66
$optimized = $uuid1->getBytes();
echo bin2hex($optimized) . "\n"; // i.e. 11d8eebc58e0a7d796690800200c9a66
echo Uuid::fromBytes($optimized)->toString() . "\n"; // i.e. 58e0a7d7-eebc-11d8-9669-0800200c9a66

// Generate multiple uuid in incrementing order
for ($i=0;$i < 10; $i++) {
    echo bin2hex(Uuid::uuid1()->getBytes()) . "\n";
}
```

Example output:
```
379ed1f8-0cac-11e6-bf97-bc5ff48511f3
379ed1f80cac11e6bf97bc5ff48511f3

379ee378-0cac-11e6-a807-bc5ff48511f3
11e60cac379ee378a807bc5ff48511f3
379ee378-0cac-11e6-a807-bc5ff48511f3

11e60cac379ee6c0a991bc5ff48511f3
11e60cac379ee7c4afe6bc5ff48511f3
11e60cac379ee8aabc1abc5ff48511f3
...
```

### DIFF
--- a/src/Codec/OrderedTimeCodec.php
+++ b/src/Codec/OrderedTimeCodec.php
@@ -1,0 +1,67 @@
+<?php
+/**
+ * This file is part of the ramsey/uuid library
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ *
+ * @copyright Copyright (c) Ben Ramsey <ben@benramsey.com>
+ * @license http://opensource.org/licenses/MIT MIT
+ * @link https://benramsey.com/projects/ramsey-uuid/ Documentation
+ * @link https://packagist.org/packages/ramsey/uuid Packagist
+ * @link https://github.com/ramsey/uuid GitHub
+ */
+namespace Ramsey\Uuid\Codec;
+
+use InvalidArgumentException;
+use Ramsey\Uuid\UuidInterface;
+
+/**
+ * OrderedTimeCodec optimizes the bytes to increment UUIDs when time goes by, to improve database INSERTs.
+ * The string value will be unchanged from StringCodec. Only works for UUID type 1.
+ */
+class OrderedTimeCodec extends StringCodec
+{
+
+    /**
+     * Encodes a UuidInterface as an optimized binary representation of a UUID
+     *
+     * @param UuidInterface $uuid
+     * @return string Binary string representation of a UUID
+     */
+    public function encodeBinary(UuidInterface $uuid)
+    {
+        $fields = $uuid->getFieldsHex();
+
+        $optimized = [
+            $fields['time_hi_and_version'],
+            $fields['time_mid'],
+            $fields['time_low'],
+            $fields['clock_seq_hi_and_reserved'],
+            $fields['clock_seq_low'],
+            $fields['node'],
+        ];
+
+        return hex2bin(implode('', $optimized));
+    }
+
+    /**
+     * Decodes an optimized binary representation of a UUID into a UuidInterface object instance
+     *
+     * @param string $bytes
+     * @return UuidInterface
+     */
+    public function decodeBytes($bytes)
+    {
+        if (strlen($bytes) !== 16) {
+            throw new InvalidArgumentException('$bytes string should contain 16 characters.');
+        }
+
+        $hex = unpack('H*', $bytes)[1];
+
+        // Rearrange the fields to their original order
+        $hex = substr($hex, 8, 4) . substr($hex, 12, 4) . substr($hex, 4, 4) . substr($hex, 0, 4) . substr($hex, 16);
+
+        return $this->decode($hex);
+    }
+}
--- a/tests/src/Codec/OrderedTimeCodecTest.php
+++ b/tests/src/Codec/OrderedTimeCodecTest.php
@@ -84,6 +84,15 @@ class OrderedTimeCodecTest extends TestCase
         $this->assertEquals($expected, $result);
     }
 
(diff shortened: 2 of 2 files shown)


ramsey:
At a glance, this looks great. Coverage decreased very slightly because that exception wasn't tested. Can you add a test to cover the exception?

barryvdh:
Added the test

barryvdh:
But please do note that I haven tested this in production, or performd any benchmarks, this is solely based on the Percona blog.

ramsey:
Thanks. I want to perform some benchmarks first, using the same methodology as that in the Percona blog post. I'll try to get to that this weekend.

barryvdh:
Did you ever get around trying this out by any chance?

ramsey:
I haven't yet. I think the Percona blog post is missing some information I need in order to reproduce the benchmarks, unless I'm missing it somewhere in the post. I'm trying to track down that information.

jmauerhan:
:+1: I have no knowledge or opinion on the InnoDB optimization, I am just super excited to see isolated contract/collaboration unit tests. Weeee!!

ramsey:
At Laracon, @jeremeamia told me that he's using this in production, and it's working well. I'd still like to see some before-and-after benchmarks, but it's time to merge this in and get folks using it. Thanks so much for contributing this!

brendt:
This approach does have a performance benefit for larger datasets: 

Comparing different methods

This data comes from a benchmark run for our Laravel package integrating this functionality: https://github.com/spatie/laravel-binary-uuid

All the benchmark data can be found here: https://docs.google.com/spreadsheets/d/1TtrHxaCdKIxQBt6jLeKZ5jI6d9QgauEc8VtXuj1SbrY/edit#gid=1559190617

alturic:
I think I'm following this correctly (I'm using TimestampFirstCombCodec however), and I can insert and select results, but I'm confused on if I'm doing it correctly, using MySQL 5.7.

For example, I'm storing the uuid's as binary(16), and if I try doing:

```
$uuidString = bin2hex(Uuid::fromString('fe15f92e-4001-11e9-b7a7-bd6b221a023a')->getBytes())

SELECT * FROM test WHERE uuid = $uuidString
```

I get no results. If I do:

```
$uuidString = bin2hex(Uuid::fromString('fe15f92e-4001-11e9-b7a7-bd6b221a023a')->getBytes())

SELECT * FROM test WHERE uuid = unhex($uuidString)
```

I get no results. If I do:

```
$uuidString = strtoupper(bin2hex(Uuid::fromString('fe15f92e-4001-11e9-b7a7-bd6b221a023a')->getBytes()))

SELECT * FROM test WHERE uuid = unhex($uuidString)
```

I do get the expected result. Is this the proper way to work with uuid's?

PR #120, opened 2016-06-16
Transition Uuid::isValid into a Validator and a provided ValidatorInterface

bradynpoulsen:
Replaces pull request #99 to prepare for release with 4.0.0


bradynpoulsen:
@ramsey coveralls is failing with the stuff that got pulled in from master/4.x

ramsey:
Coveralls isn't "failing." It's just showing a decrease in coverage, which is acceptable. :-)

ramsey:
Thanks for all this, @bradynpoulsen! I'll re-review it this weekend and get it merged in.

src/Uuid.php
```diff
@@ -93,7 +94,7 @@ class Uuid implements UuidInterface
     /**
      * Regular expression pattern for matching a valid UUID of any variant.
      */
-    const VALID_PATTERN = '^[0-9A-Fa-f]{8}-[0-9A-Fa-f]{4}-[0-9A-Fa-f]{4}-[0-9A-Fa-f]{4}-[0-9A-Fa-f]{12}$';
+    const VALID_PATTERN = Validator::VALID_PATTERN;
```
bradynpoulsen: @ramsey Since this will be released with 4.0, should we remove this and just have it live on the validator vs UUID as well?

ramsey: That's probably a good idea.

src/UuidFactory.php
```diff
@@ -180,6 +187,24 @@ public function setUuidBuilder(UuidBuilderInterface $builder)
...
+    {
+        return $this->validator;
+    }
+
+    /**
+     * @return void
+     */
+    public function setValidator(ValidatorInterface $validator)
+    {
+        assert(func_num_args() === 1);
```
aztech-dev: Not sure this is really necessary. If you have less than one arg, the call will fail as $validator cannot be null, and if there are more than one, well they'll just be ignored.

bradynpoulsen: True, I guess it just depends on how strong you like method signatures to be. I'll remove this since it's just a library though

bradynpoulsen:
@ramsey It's failing because of a missing function `iconv_strlen`?

ramsey:
That looks like an error in something else that's trying to invoke a polyfill function that doesn't exist. I'll try digging into it to see what's going on.

moderndeveloperllc:
@ramsey Are you using custom docker files for your Travis runs? It would be odd that the regular Travis PHP would not have iconv installed as it's a default installed extension.

ramsey:
@moderndeveloperllc I am using custom Docker images for the builds. This is so that I can test the library on 32-bit and big-endian systems. See here: https://github.com/ramsey/uuid/blob/master/util/build-docker-image.sh

ramsey:
I'm confused, though, since I haven't changed anything in that Docker image in months, and the tests only just started failing for this PR.

moderndeveloperllc:
`./configure --disable-all` will definitely keep iconv and mbstring from being compiled into PHP, so it's not odd that you would get that error. [symfony/console](https://github.com/symfony/console/blob/master/composer.json#L20) is requiring `symfony/polyfill-mbstring`, but not `symfony/polyfill-iconv`. I attempted to raise this as [an issue](https://github.com/symfony/polyfill/issues/66), but was told they would not require it.

What is odd is that you are explicitly calling `strlen()` and it's running the polyfill. I can't tell if it's Symfony, or the docker image that is overloading the multibyte functions.

ramsey:
Please merge the 4.x branch back into this and resolve the conflicts. I've reverted to running the tests directly using Travis's default PHP instead of a Docker container, so the `iconv_strlen` issue should no longer be a problem.

Thanks!

ramsey:
I'm so sorry for taking so long to merge this. I thought I had merge it ages ago. Merging in now. Thanks!

PR #127, opened 2016-06-23
Replace moontoast/math by GMP

### LINKED ISSUE #109: Use a math library other than moontoast/math
A math library is not required for this library unless users want to operate on the UUID as a 128-bit integer. For this purpose, the library has suggested moontoast/math as an optional dependency, and internally, this is what ramsey/uuid uses, if it is present. However, moontoast/math is no longer maintained.

I will either create a ramsey/math library that allows the use of different adapters for working with large numbers, or I want to find an existing math library that provides similar functionality.

jecknig:
This resolves #109

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # ramsey/uuid Changelog
 
+## 3.4.1
+
+_Released: 2016-04-23_
+
+  * Fixed test that violated a PHP CodeSniffer rule, breaking the build.
+
 ## 3.4.0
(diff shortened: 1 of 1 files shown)


ramsey:
This is all looking good, but I think we should keep the `BigNumberConverter` class for now and introduce a new `GmpConverter` class that does what you've implemented here. "BigNumber" is the name of the class from the moontoast/math library. Then, resubmit the pull request to the 4.x branch, please.

Thanks!

PR #128, opened 2016-07-03
Add GmpConverter and GmpTimeConverter

### LINKED ISSUE #109: Use a math library other than moontoast/math

jecknig:
This resolves #109.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ After doing so, you will have the latest ramsey/uuid package in the 2.x series, 
 
 ## Requirements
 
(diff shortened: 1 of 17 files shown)


ramsey:
This is looking great. At a glance, the only thing that concerns me is the change to the Uuid constructor. The addition of required parameters could be problematic, but as this is for the 4.x branch, maybe that's not such a big deal, though I'm considering back-porting much of this to the 3.x branch and deprecating the use of moontoast/math.

I'm going to review this in more detail later this week. Thanks!

ramsey:
Merging into 4.x. Thanks!

PR #129, opened 2016-07-28
Fix RandomNodeProvider: change the max value in mt_rand() call to prevent invalid node generation

Antoha-GS:
When at least one call of `mt_rand(0, 1 << 24)` in  `RandomNodeProvider::getNode()` returns `1 << 24 // 0x1000000` the method generates a hexadecimal string that contain one digit more than needed to be correct. As a result, exception 'Invalid node value' is thrown when we trying to generate UUIDv1 with `DefaultTimeGenerator`. I faced with it in practice.

Therefore, i suggest to use `0xffffff` instead of `1 << 24` to fix this behaviour.

### DIFF
--- a/src/Provider/Node/RandomNodeProvider.php
+++ b/src/Provider/Node/RandomNodeProvider.php
@@ -31,6 +31,6 @@ class RandomNodeProvider implements NodeProviderInterface
      */
     public function getNode()
     {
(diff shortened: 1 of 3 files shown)


ramsey:
Thanks for fixing this!

PR #130, opened 2016-08-01
Fixed undefined classes in GuidStringCodecTest properties

ondrejmirtes:
I found out these two phpDocs pointed to undefined classes. I added them as use statements.

BTW: I found this issue using my new static analysis tool [PHPStan](https://github.com/phpstan/phpstan). I can send you a PR with integrating the tool in the package's build script if you're interested :)

### DIFF
--- a/tests/src/Codec/GuidStringCodecTest.php
+++ b/tests/src/Codec/GuidStringCodecTest.php
@@ -2,8 +2,10 @@
 
 namespace Ramsey\Uuid\Test\Codec;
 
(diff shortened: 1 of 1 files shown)


ramsey:
Thanks!

We can't use PHPStan with this project, since we support 5.4+, and the next major version will support 5.6+.

ondrejmirtes:
You're welcome.

It'd still be possible to run it under the PHP 7-build only and have the build matrix fail if there was something wrong with the code (and just let the build step pass on lower versions).

PR #134, opened 2016-09-08
Fix: Reduce visibility of setUp() and tearDown()

localheinz:
This PR

* [x] reduces the visibility of `seUp()` and `tearDown()` from `public` to `protected`

💁 This is the visibility as declared on the parent; there's no need to increase it.


ramsey:
Thanks!

PR #136, opened 2016-09-15
Fix: Move tests one up

localheinz:
This PR

* [x] moves tests one level up

💁 Not sure why, but it's the first time ever that I have seen tests reside in `tests/src` instead of `tests` or `test`.


ramsey:
Last year, I made this commit in which I restructured the tests, introducing the `tests/src/` directory: f5e96b6af779a1c9fb7cc3acc012215bfcb66db1

Honestly, I don't remember why I introduced it, but looking at all my other projects, I believe I was aiming to be consistent in my project structure. I do the same in ramsey/collection, ramsey/math, bitworking/mimeparse, and moontoast/math. It's likely that I provided an extra level of depth to separate `src/bootstrap.php` from the folder containing all the test files (ramsey/uuid doesn't use a bootstrap.php file, while other projects I work on do). I've also noted a number of other PHP library projects who follow this convention, so I was aiming to conform to what I saw other projects doing.

I don't disagree with your pull request; in fact, I have no opinion either way (to keep or get rid of the extra directory level under `tests/`), but I'm inclined to leave things the way they are, rather than introduce another massive restructuring of the tests.

I'm willing to be convinced otherwise, though. :-)

jmauerhan:
It's definitely not the norm, but with the namespacing it really doesn't matter, it has no real impact on the code, so I don't see a benefit in moving them other than personal preferences.

localheinz:
@jmauerhan ramsey 

The namespace definitely makes sense, doing the same thing in all of my projects:

```json
{
    "autoload": {
        "psr-4": {
            "Foo\\Bar\\": "src"
        }
    },
    "autoload-dev": {
        "psr-4": {
            "Foo\\Bar\\Test\\": "test"
        }
    }
}
```

The structure above also makes it easy to split tests, for example

```
test
├── Acceptance
├── Integration
└── Unit
```

without a need to adjust any autoloading configuration.

But I'm fine with whatever makes the most sense in the context of this project, so feel free to close or merge at will!

ramsey:
After thinking about it some more, I'll merge it in. Thanks!

localheinz:
@ramsey 

👍

Issue #131, opened 2016-08-02
UUID in lower and uppercase do not equal while they should


sstok:
When you take the string representation of a UUID and convert it to uppercase it no longer equals when comparing, this is caused by the fact that `compareTo()` only compares them as characters and A and a are not the same.

```php
<?php

use Ramsey\Uuid\Uuid;

require 'vendor/autoload.php';

$uuid = Uuid::uuid4();

$uuidString = (string) $uuid;
$upperUuid = strtoupper($uuidString);

echo sprintf("normal: '%s'; equals: %s\n", $uuidString, var_export(Uuid::fromString($uuidString)->equals($uuid), true));
echo sprintf("uppercase: '%s'; equals: %s\n", $upperUuid, var_export(Uuid::fromString($upperUuid)->equals($uuid), true));
```

Should the casing be normalized in fromString or something, or is this something the implementer should take care of?

ramsey:
Good catch. It is comparing the hex values for equality. We'll need to do a conversion to either lowercase or uppercase when setting the fields on the UUID, so I'll mark this issue as a bug to fix.

Thanks!

jmauerhan:
@ramsey have you planned any work on this yet? I'd be interested in working on it if not.

ramsey:
Go for it!

PR #137, opened 2016-10-02
UUID equals case-insensitive

jmauerhan:
Fixes https://github.com/ramsey/uuid/issues/131

- Added test cases for equals, compareTo, and fromString. 
- Added strtolower in fromString so all UUID created from a string will be converted into lowercase.


ramsey:
Looks great. Thanks, and happy Hacktoberfest! :-)

PR #140, opened 2016-10-26
Clean up more repo meta files

afurculita:
This PR removes the Changelog, Code of conduct and Contributing files at export.

Also mark all files for automatic end-of-line normalization.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,8 +1,13 @@
-/.gitignore       export-ignore
-/.gitattributes   export-ignore
-/.travis.yml      export-ignore
-/.scrutinizer.yml export-ignore
-/tests            export-ignore
-/phpunit.xml.dist export-ignore
-/phpcs.xml        export-ignore
-/util             export-ignore
+* text=auto
+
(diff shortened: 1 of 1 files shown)


ramsey:
I'm happy to include the `* text=auto` addition to the `.gitignore` file, but the changelog, code of conduct, and contributing files should be part of all distributions.

Thanks!

afurculita:
I find relevant for distributions only the README and LICENSE files. An example of popular distributions that don't include other files is Laravel, flysystem.

ramsey:
Laravel does not have a changelog, code of conduct, or contributing file in their repo. Flysystem does not have code of conduct or contributing files.That's why they don't include these files in their distributions.

These files are not simply code/repository metadata. They are as important and relevant to the distribution as the README and LICENSE files. If another library/project maintains these files in their repo and excludes them from the distribution, it my opinion that they are in error.

Other package maintainers are welcome to disagree, but these files will remain in my distributions. :-)

That said, I do appreciate your efforts to help improve this library by cleaning up things. Thanks!

PR #143, opened 2016-10-31
Update AOP framework to latest 1.0 version

lt:
I was getting a lot of errors when running tests.

```
PHPUnit_Framework_Exception: PHP Notice:  Undefined index: Doctrine\Common\Annotations\ in /home/leigh/uuid/vendor/goaop/framework/src/Instrument/ClassLoading/AopComposerLoader.php on line 78
```

This seems to have been fixed since the alpha version of Go! AOP, and all tests pass after upgrading to the latest version (1.0.2).


lt:
Oh, failing on PHP version. Damnit!

PR #144, opened 2016-10-31
Pin doctrine/annotations to a 1.2.x version (Fixes build)

lt:
1.3.x is currently causing test failures when used with goaop
1.0.0-alpha2 (which we are pinned at for PHP 5.4 compat)


ramsey:
Thanks so much!

PR #146, opened 2016-11-30
Typo

grogy:
It is too too small change, but for read it is better :-)

_A dot pass after the sentence, not in the cite._

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ composer require ramsey/uuid
 
 While we have made significant internal changes to the library, we have made every effort to ensure a seamless upgrade path from the 2.x series of this library to 3.x.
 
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
I don't think this is correct. The period was intended to be inside the quote because it belongs to the sentence inside the quote.

grogy:
Thanks for your comment, [I looked to grammar](http://www.grammarbook.com/punctuation/quotes.asp) it now. My mistake, I take it in my grammar (czech), sorry.

ramsey:
In American English grammar, punctuation always goes inside quotation marks. It can seem counterintuitive, I know, but such are our grammar rules. :-)

PR #158, opened 2017-03-08
Be more specific about PHP versions

Nyholm:
We do not support PHP6 nor PHP8.


ramsey:
Thanks!

Nyholm:
Thank you for merging

PR #160, opened 2017-03-09
Improves memory usage when generating large quantities of UUID's

iainmckay:
Improves memory usage in `UuidFactory::uuidFromHashedName` when generating large volumes of UUID's as documented in issue #159.

PR #164, opened 2017-03-26
Optimize UUID string decoding

mhujer:
I realized that 30% of the request time in our app is spent by hydrating the uuid (We are using Doctrine and in this specific request I was accidentally hydrating few thousands entities).

Luckily, I remembered the #160 and tried to do similar optimization for uuid decoding. It resulted in 10-20% performance improvement (it is not much, but every millisecond counts).

Benchmark: last commit in https://github.com/mhujer/uuid/commits/mh-uuid-decoding-speedup-benchmark
Benchmark results (Blackfire): https://blackfire.io/profiles/compare/42276a3a-cb45-47e2-80c4-61420b3c61bd/graph

### DIFF
--- a/src/Codec/StringCodec.php
+++ b/src/Codec/StringCodec.php
@@ -156,12 +156,12 @@ class StringCodec implements CodecInterface
     protected function getFields(array $components)
     {
         return array(
-            'time_low' => sprintf('%08s', $components[0]),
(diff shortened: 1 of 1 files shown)


ramsey:
Good catch! Thanks for contributing.

mhujer:
Thanks for quick merge and release! 👍

PR #168, opened 2017-04-27
allow valid uuid hex strings to pass isValid method

charliekassel:
If a hex string can be converted to a Uuid without exception, it should be valid.

### DIFF
--- a/src/Uuid.php
+++ b/src/Uuid.php
@@ -629,11 +629,34 @@ class Uuid implements UuidInterface
             return true;
         }
 
+        if (self::looksLikeHex($uuid)) {
+            try {
+                $uuid = self::fromString($uuid);
+            } catch (\InvalidArgumentException $e) {
+                return false;
+            }
+        }
+
         if (!preg_match('/' . self::VALID_PATTERN . '/', $uuid)) {
             return false;
         }
 
         return true;
+    }
+
+    /**
+     * Simple test to see if string could be a 32 char hex
+     * 
+     * @param  string $uuid The string UUID to test
+     * @return boolean
+     */
+    private static function looksLikeHex($uuid)
+    {
+        if (strlen($uuid) === 32) {
+            return true;
+        }
+
+        return false;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


ramsey:
The point of the `testIsValidNoDashes()` is to assert that it is `false`, since a UUID without dashes is not a valid string UUID. I don’t want to use `fromString()` in the `isValid()` check because of the potential performance penalty involved. Is there a reason you are unable to do this same check in your application?

charliekassel:
Right ok, so a hex string is not a valid uuid but it can represent one?

ramsey:
A 32-character hexadecimal string is not a valid representation of a UUID. See here: https://tools.ietf.org/html/rfc4122#section-3

In particular:

```
      The formal definition of the UUID string representation is
      provided by the following ABNF [7]:

      UUID                   = time-low "-" time-mid "-"
                               time-high-and-version "-"
                               clock-seq-and-reserved
                               clock-seq-low "-" node
      time-low               = 4hexOctet
      time-mid               = 2hexOctet
      time-high-and-version  = 2hexOctet
      clock-seq-and-reserved = hexOctet
      clock-seq-low          = hexOctet
      node                   = 6hexOctet
      hexOctet               = hexDigit hexDigit
      hexDigit =
            "0" / "1" / "2" / "3" / "4" / "5" / "6" / "7" / "8" / "9" /
            "a" / "b" / "c" / "d" / "e" / "f" /
            "A" / "B" / "C" / "D" / "E" / "F"

   The following is an example of the string representation of a UUID as
   a URN:

   urn:uuid:f81d4fae-7dec-11d0-a765-00a0c91e6bf6
```

`Uuid::getHex()` is provided not as a means to strip the dashes from the UUID but as a means to return a string (hexadecimal) representation of the 128-bit integer that makes up a UUID.

My recommendation is to always store UUIDs in binary format (16 char). Using `Uuid::fromString()` and `Uuid::fromBytes()` are the best ways to determine whether the string or bytes actually does represent a UUID.

I do appreciate your time in putting together this PR. Based on the UUID specification, I disagree that this library should validate 32-character hexadecimal strings as UUIDs.

charliekassel:
No worries, many thanks anyway for the library.
I realise that i'm just hacking at it and not fully understanding UUID, so I respect your judgement on this.

ramsey:
Thanks for hacking. Keep it up!

PR #170, opened 2017-05-18
Multicast Bit not Set

crazytonyi:
Per RFC 4122, 4.6:

>  For systems with no IEEE address, a randomly or pseudo-randomly generated value may be used; see Section 4.5.  The multicast bit must be set in such addresses, in order that they will never conflict with addresses obtained from network cards.

and 4.5:

> A better solution is to obtain a 47-bit cryptographic quality random number and use it as the low 47 bits of the node ID, with the least significant bit of the first octet of the node ID set to one.  This bit is the unicast/multicast bit, which will never be set in IEEE 802 addresses obtained from network cards.  Hence, there can never be a conflict between UUIDs generated by machines with and without network cards.  (Recall that the IEEE 802 spec talks about transmission order, which is the opposite of the in-memory representation that is discussed in this document.)

> For compatibility with earlier specifications, note that this document uses the unicast/multicast bit, instead of the arguably more correct local/global bit.

Thus, the Least Significant Bit of the randomly-generated node ID should be set to 1 to indicate that it is a randomly-generated node ID.

Proposed change is to flip the least-significant bit of the second randomly-generated number via inclusive OR bitwise operator.

### DIFF
--- a/src/Provider/Node/RandomNodeProvider.php
+++ b/src/Provider/Node/RandomNodeProvider.php
@@ -31,6 +31,6 @@ class RandomNodeProvider implements NodeProviderInterface
      */
     public function getNode()
     {
-        return sprintf('%06x%06x', mt_rand(0, 0xffffff), mt_rand(0, 0xffffff));
+        return sprintf('%06x%06x', mt_rand(0, 0xffffff), mt_rand(0, 0xffffff) | 0x01);
     }
 }


ramsey:
Addresses issue #171

jobe1986:
According to the following reference links this patch appears to set the WRONG bit to 1 for a multicast IEEE address. The quoted text in the initial post even states "the first octet" not the last where as the patch applies the bit change to the last octet. Also comparing with the output of the command "uuid -mv 1" also shows that it is the least significant bit of the first octet that is set to 1.

References:
https://en.wikipedia.org/wiki/MAC_address#Unicast_vs._multicast
http://bradhedlund.com/2007/11/21/identifying-ethernet-multicast/
https://technet.microsoft.com/en-us/library/cc957928.aspx

Issue #173, opened 2017-07-30
Exposing versions constants or enum


davispeixoto:
Hello @ramsey , I'd like to add an improvements suggestion for this amazing package.

It would be great if you can expose constants for algorithm versions in your package (the same as you use on the PeclUuidTestCase - eg, UUID_TYPE_TIME and UUID_TYPE_RANDOM.

My suggestion is to expose constants or an enum (using something like myclabs/enum package).

This would help us avoiding magic numbers in cases where we would like to enforce a certain version of UUID. Suppose I wanna enforce all my ids to be UUID4.

Right now it require me to enforce it using magic numbers, or implementing constants into my own application, or implemeting an enumeration with them.

Including this can make your package more self contained for that kind of situation.

Thank you in advance.

davispeixoto:
Hey @ramsey , thanks for replying so promptly.

I'm writing a tutorial about architecture best practices. One of the requests is to use UUID and make sure it is a UUID4 type for all of our entities. Cuurently I have this method implemented on our AbstractEntity class:

```
  const UUID_TYPE_RANDOM = 4;
...
    /**
     * @param UuidInterface $uuid
     * @return UuidInterface
     * @throws InvalidUuidStringException
     */
    private function isVersionValid(UuidInterface $uuid)
    {
        if ($uuid->getVersion() !== self::UUID_TYPE_RANDOM) {
            throw new InvalidUuidStringException('Not supported version: '.$uuid->getVersion());
        }

        return $uuid;
    }
```

So if a UUID5 for example is passed, the entity itself won't accept it.

I had an issue while writing this before, because PHPStorm used to highlight a constant with that name (UUID_TYPE_RANDOM), luckly my tests failed soon, and I realized that constant is valid only within a test. This lead me to suggest this implementation.

davispeixoto:
Hi. I've added 3 different ways to do this as pull requests.

ramsey:
Thanks for your help implementing this! I've merged #177 into master and have tagged the 3.7.0 release with these changes.

davispeixoto:
I thank you for accepting my contribution. I've learned a lot with this well curated and tested package.
Definitively we need more like these into php community.

PR #175, opened 2017-08-03
issue #173 expose versions alternative3

davispeixoto:
I suggested an enhancement, but after reading the contribution guide I've decided to implement it myself in three flavors (so you can decide which is better, and eventually merge). All of them are tested.

This one is an Enum implementation.

Thx


ramsey:
Closing in favor of #177.

PR #177, opened 2017-08-03
issue #173 expose versions alternative1

davispeixoto:
I suggested an enhancement, but after reading the contribution guide I've decided to implement it myself in three flavors (so you can decide which is better, and eventually merge). All of them are tested.

This one is perhaps the most simple one. I've just added 5 constants into Uuid class, and tests for them.

Thx

### DIFF
--- a/tests/UuidTest.php
+++ b/tests/UuidTest.php
@@ -1968,4 +1968,49 @@ class UuidTest extends TestCase
 
         $this->assertEquals('b6c54489-38a0-5f50-a60a-fd8d76219cae', $uuid->toString());
     }
+
+    /**
+     * @depends testGetVersionForVersion1
+     */
+    public function testUuidVersionConstantForVersion1()
+    {
+        $uuid = Uuid::fromString('ff6f8cb0-c57d-11e1-9b21-0800200c9a66');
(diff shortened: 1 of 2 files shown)


tests/UuidTest.php
```diff
@@ -1968,4 +1968,49 @@ public function testUuid5WithZeroName()
...
         $this->assertEquals('b6c54489-38a0-5f50-a60a-fd8d76219cae', $uuid->toString());
     }
+
+    /**
+     * @depends testGetVersionForVersion1
+     */
+    public function testUuidVersionConstantForVersion1()
+    {
+        $uuid = Uuid::fromString('ff6f8cb0-c57d-11e1-9b21-0800200c9a66');
+        $this->assertEquals($uuid->getVersion(), UUID::UUID_TYPE_TIME);
```
ramsey: While it technically passes, please update the class name on the constant here to use the proper capitalization: `Uuid` instead of `UUID`. All the tests below this will also need this change. So, in this example, it should be `Uuid::UUID_TYPE_TIME`. Thanks!

davispeixoto:
Change request is done

davispeixoto:
Thx for accepting my contribution. I've learned a lot with this well curated and tested package.

Issue #53, opened 2015-03-25
how to generate comb uuid?


Daskul:
I've tried to generate atleast thousands of uuid using Uuid::uuid4() but they are not in sequence.

aztech-dev:
Current release does not provide COMB support. You'll have to wait for the 3.0 release to generate COMB's.

Daskul:
I see. I thought I've read before that it was already supported. Thank you for the information

hi2u:
Now that it's in the library, it would be good if it was documented in the README.md 

I couldn't figure it out until I came across https://benramsey.com/blog/2016/04/ramsey-uuid/

```php
$factory = new \Ramsey\Uuid\UuidFactory();
$generator = new \Ramsey\Uuid\Generator\CombGenerator($factory->getRandomGenerator(), $factory->getNumberConverter());
$codec = new \Ramsey\Uuid\Codec\TimestampFirstCombCodec($factory->getUuidBuilder());
 
$factory->setRandomGenerator($generator);
$factory->setCodec($codec);
 
\Ramsey\Uuid\Uuid::setFactory($factory);
 
$uuid = \Ramsey\Uuid\Uuid::uuid4();
```

Is this the easiest way to do it?  

Would it also be worth adding a simple method to the main class such as Uuid::uuid4comb() that does it for us without pasting all the above into our own code?

ramsey:
Thanks for the recommendation. I'm currently working on documentation for this and hope to have it posted soon.

1ma:
How feasible would it be to add a new static method to `Uuid` to generate this kind of UUIDs?

ramsey:
@1ma That sounds like it could be reasonable to me. Something like `Uuid::comb()`?

ramsey:
Sounds great. Thanks!

kael-shipman:
Just a reminder: Looks like this never made it into the docs. I'm using @hi2u's implementation for now but would love to know if there's a "standard" way to do it.

broofa:
FWIW, we (node-uuid) decided to pass on supporting "comb" uuids.  See https://github.com/kelektiv/node-uuid/issues/303#issuecomment-458757302

tl;dr: RFC4122 Version 1 UUIDs serve basically the same purpose.

hi2u:
@broofa The difference is that UUIDv1 isn't actually in timestamp order. 

[...]

PR #180, opened 2017-09-03
added helper method for COMB UUIDs

1ma:
This PR add a helper method to the `Uuid` class for generating sequential COMB UUIDs, as discussed on issue #53.

Disclaimer: this proposal is a WIP. Though the feature seems to work as intended I've decided to open the PR early on to get feedback on the approach (feels a bit clunky to me). I'd also like input on how to approach testing such feature.

### DIFF
--- a/src/UuidFactory.php
+++ b/src/UuidFactory.php
@@ -14,7 +14,9 @@
 
 namespace Ramsey\Uuid;
 
+use Ramsey\Uuid\Codec\TimestampFirstCombCodec;
 use Ramsey\Uuid\Converter\NumberConverterInterface;
+use Ramsey\Uuid\Generator\CombGenerator;
 use Ramsey\Uuid\Provider\NodeProviderInterface;
 use Ramsey\Uuid\Generator\RandomGeneratorInterface;
 use Ramsey\Uuid\Generator\TimeGeneratorInterface;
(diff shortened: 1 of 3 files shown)


ramsey:
At first glance, the approach looks good to me. I can take a more detailed look later this week and offer some pointers on testing. Thanks!

src/UuidFactory.php
```diff
@@ -229,6 +231,23 @@ public function uuid5($ns, $name)
         return $this->uuidFromNsAndName($ns, $name, 5, 'sha1');
     }
 
+    public function comb()
+    {
+        $currentCodec = $this->getCodec();
+        $combGenerator = new CombGenerator($this->getRandomGenerator(), $this->getNumberConverter());
+        $combCodec = new TimestampFirstCombCodec($this->getUuidBuilder());
```
ramsey: There's a difference between a COMB UUID and a Timestamp First COMB UUID. I'm worried that that name of the method `comb()` causes confusion, since you're using the `TimestampFirstCombCodec()` internally.

1ma: Indeed but `TimestampFirstCombCodec` is the one that produces monotonically increasing UUIDs, which is the whole point.

Do you have an alternative name in mind?

ramsey: Maybe `timestampFirstComb()`?

ramsey:
Any other thoughts on choosing a different name? I still disagree with using `comb()`, since it's not a standard COMB UUID but is a timestamp-first COMB UUID.

1ma:
It's ok with me. It appears that I managed to nuke my fork though, sorry. I'll resubmit the PR (with the name of the helper set to `timestampFirstComb()`)

PR #182, opened 2017-09-22
Set the multicast bit for random nodes, according to RFC 4122, §4.5

### LINKED ISSUE #171: Multicast bit not set for randomly-generated network nodes in v1
Not sure how to tie an issue to a PR, but figured it is preferred that peanut-gallery PRs have a related issue. As noted in PR #170:

Per RFC 4122, 4.6:

>  For systems with no IEEE address, a randomly or pseudo-randomly generated value may be used; see Section 4.5.  The multicast bit must be set in such addresses, in order that they will never conflict with addresses obtained from network cards.

and 4.5:

> A better solution is to obtain a 47-bit cryptographic quality random number and use it as the low 47 bits of the node ID, with the least significant bit of the first octet of the node ID set to one.  This bit is the unicast/multicast bit, which will never be set in IEEE 802 addresses obtained from network cards.  Hence, there can never be a conflict between UUIDs generated by machines with and without network cards.  (Recall that the IEEE 802 spec talks about transmission order, which is the opposite of the in-memory representation that is discussed in this document.)

> For compatibility with earlier specifications, note that this document uses the unicast/multicast bit, instead of the arguably more correct local/global bit.

Thus, the Least Significant Bit of the randomly-generated node ID should be set to 1 to indicate that it is a randomly-generated node ID.

Proposed change is to flip the least-significant bit of the second randomly-generated number via inclusive OR bitwise operator.

ramsey:
When merged, this will close #171 and close #170.

### DIFF
--- a/src/Provider/Node/RandomNodeProvider.php
+++ b/src/Provider/Node/RandomNodeProvider.php
@@ -31,6 +31,11 @@ class RandomNodeProvider implements NodeProviderInterface
      */
     public function getNode()
     {
-        return bin2hex(random_bytes(6));
+        $node = hexdec(bin2hex(random_bytes(6)));
+
+        // Set the multicast bit; see RFC 4122, section 4.5.
+        $node = $node | 0x010000000000;
+
+        return dechex($node);
     }
 }
--- a/src/Provider/Node/RandomNodeProvider.php
+++ b/src/Provider/Node/RandomNodeProvider.php
@@ -36,6 +36,6 @@ class RandomNodeProvider implements NodeProviderInterface
         // Set the multicast bit; see RFC 4122, section 4.5.
         $node = $node | 0x010000000000;
 
-        return sprintf('%012x', $node);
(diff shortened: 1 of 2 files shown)


ramsey:
@jobe1986 and @crazytonyi, would you like to review this before I merge it to master? This addresses the multicast bit being set on the wrong bit in #170. I'll be merging it in soon, since I'd like to get this out in a release. Thanks!

jobe1986:
Just one thing to note, use of random_bytes() will increase the required PHP version to 7 as said function is not available in prior versions. But apart from that the code in question appears to function correctly.

ramsey:
@jobe1986 I am using [paragonie/random_compat](https://packagist.org/packages/paragonie/random_compat) as a polyfill, and all tests pass on PHP 5.4, 5.5, 5.6, 7.0, 7.1, and the nightlies. :smile:

Thanks for reviewing!

ramsey:
Some of the Travis CI jobs are timing out, but the build is otherwise passing on all supported versions of PHP, so I'm merging this into master.

PR #185, opened 2017-10-07
Add feature to use sysfs interface to get node identifier

### LINKED ISSUE #184: Uuid1 class uuids have a random node identifier in minimal systems
This is because the netstat binary not being installed in minimal linux systems, and it's a legacy tool so the installation base is expected to shrink.

Calling the following script in a docker container like php:7
```php
<?php
require 'vendor/autoload.php';

use Ramsey\Uuid\Uuid;

$uuid1 = Uuid::uuid1();
echo $uuid1->getNodeHex() ."\n";
```                   

results in
```
root@e675ab19962f:/code# php foo.php 
cd9947033709
root@e675ab19962f:/code# php foo.php 
4256b01c8d70
```

While the expected result is
```
root@e675ab19962f:/code# php foo.php
0242ac110003
root@e675ab19962f:/code# php foo.php 
0242ac110003
```
where 0242ac110003 is the mac address  of the network interface inside the docker container

mjrider:
Linux offers a sysfs interface to the mac addresses of the network
interfaces without calling exec. this fixes #184

This enables repeatable system based UUID generation even on docker images
like php:7 

TODO:
- [x] fix unit testing, it fails on getIfconfig not being called, i think the way of mocking needs to be changed.
- [x] write changelog, when all changes are finalized

### DIFF
--- a/src/Provider/Node/SystemNodeProvider.php
+++ b/src/Provider/Node/SystemNodeProvider.php
@@ -39,11 +39,11 @@ class SystemNodeProvider implements NodeProviderInterface
         $matches = array();
 
         // first try a  linux specific way
-
         $node = $this->getsysfs();
+
         // Search the ifconfig output for all MAC addresses and return
         // the first one found
-        if ($node !== false) {
+        if ($node === false) {
             if (preg_match_all($pattern, $this->getIfconfig(), $matches, PREG_PATTERN_ORDER)) {
                 $node = $matches[1][0];
             }
@@ -86,6 +86,7 @@ class SystemNodeProvider implements NodeProviderInterface
      */
     protected function getsysfs()
     {
+        $mac = false;
         if (strtoupper(php_uname('s')) === "LINUX") {
             // get all the macadresses of all systems
             $macs = array_map(
--- a/src/Provider/Node/SystemNodeProvider.php
+++ b/src/Provider/Node/SystemNodeProvider.php
@@ -38,14 +38,19 @@ class SystemNodeProvider implements NodeProviderInterface
         $pattern = '/[^:]([0-9A-Fa-f]{2}([:-])[0-9A-Fa-f]{2}(\2[0-9A-Fa-f]{2}){4})[^:]/';
         $matches = array();
 
+        // first try a  linux specific way
+
+        $node = $this->getsysfs();
         // Search the ifconfig output for all MAC addresses and return
         // the first one found
-        $node = false;
-        if (preg_match_all($pattern, $this->getIfconfig(), $matches, PREG_PATTERN_ORDER)) {
-            $node = $matches[1][0];
+        if ($node !== false) {
+            if (preg_match_all($pattern, $this->getIfconfig(), $matches, PREG_PATTERN_ORDER)) {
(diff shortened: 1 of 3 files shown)


mjrider:
i have trouble understanding how the mocking of function in the unit testing works, and need some help with that to get this PR mergeable

ramsey:
@mjrider, I can provide some pointers on mocking the functions this week. Thanks!

mjrider:
@ramsey i checked previous PR's but can't find the right way to add a changelog entry for a not yet released  version, die you have a sample for that, or do you write that yourself on release.

further i think the code is function correctly, but the builds are timing out.

ramsey:
@mjrider I can write it myself. Sorry I haven't been able to get around to showing an example of how to mock the calls.

ramsey:
I've restarted the jobs. Hopefully, they won't timeout this time.

mjrider:
i tried to mock the file_get_contents call, but because it is not a function call, but a string argument, it seems not to work. 

any help in getting it to work would be appreciated

mjrider:
i think , with some help of @Potherca, i got the mocking working

mjrider:
@ramsey could you tell me if the changed made are what you wanted with your review?

src/Provider/Node/SystemNodeProvider.php
```diff
@@ -73,4 +78,35 @@ protected function getIfconfig()
...
+
+            $macs = array_map('trim', $macs);
+
+            // remove invalid entries
+            $macs = array_filter($macs, function ($mac) {
+                return
+                    // localhost adapter
+                    $mac !== '00:00:00:00:00:00' &&
+                    // must match  mac adress
+                    preg_match('/^([0-9a-f]{2}:){5}[0-9a-f]{2}$/i', $mac);
```
ramsey: Should this use the same pattern that is used in `getNode()` earlier this in this class? If so, I'd move that pattern into a class constant and use it in both places, to avoid duplication.

mjrider: no i think we should not, this patterns has an anchor to the start and the end of the line, while the other patterns requires a char, not being an `:`  on both sides of the macaddress. While both patterns look alike, the are both for different purposes and expected input, and thus subtle different.

ramsey:
Merged in. I’ll package up a release soon. Thanks for contributing!

src/Provider/Node/SystemNodeProvider.php
```diff
@@ -73,4 +78,35 @@ protected function getIfconfig()
...
+     * @return string|bool
+     */
+    protected function getsysfs()
+    {
+        $mac = false;
+        if (strtoupper(php_uname('s')) === "LINUX") {
+            // get all the macadresses of all systems
+            $macs = array_map(
+                'file_get_contents',
+                glob('/sys/class/net/*/address', GLOB_NOSORT)
```
nylen: It's been a while since this was written and merged, but this will fail on some systems and needs error handling.  Here's what I'm seeing:

> file_get_contents(/sys/class/net/eth0/address): failed to open stream: Permission denied

Issue #189, opened 2017-10-18
[Version 4.0.0] Upgrade unit-tests to PHPUnit v6


Potherca:
As discussed in #186, when `ramsey/uuid` v4 is released it would be advisable to upgrade to PHPUnit 6.4 (as support for PHPUnit 5 ends in February 2018).

In order to make tests run under PHPUnit 6, some calls will need to be changed (as PHPUnit 6 removes a lot of things that have been deprecated under PHPUnit 5).

A full list of changes across PHPUnit versions can [be found here](https://gist.github.com/Potherca/1533bde47294a2bb860b18431d2c6291).

The most important changes are:

- [x] Require PHPUnit `^6.4` (and thus PHP `^7.0`) in the `composer.json` file
- [x] `PHPUnit_Framework_TestCase` no longer exists
- [x] `TestCase::setExpectedException()` has been replaced with `TestCase::expectException()`
- [x] `TestCase::getMock()` has been replaced with `TestCase::getMockBuilder()->getMock()` or `TestCase::createMock` (fixed in #188, not yet merged to `master` or `4.x` branches).

Maybe superfluous to state, but any pull-requests for these changes should be opened against the `4.x` branch (not the `master` branch).

ethanbray:
Are you working on this personally? I would be willing to make these changes myself, just don't want to duplicate work you've already done.

Potherca:
I have not started any serious work on any of the other items. I was waiting for #188 to get merged before devoting more time to this issue.

ramsey:
I've just merged in #188 to master. Thanks for the reminder!

ramsey:
I decided to update the 4.x branch so that it only runs builds on PHP 7 and up. In the process, I bumped up all the dev packages and fixed the build failures. My changes may have checked off everything on your list.

The only thing I didn't do was use `^6.4` for PHPUnit; I used `^6.0`. If you'd like to upgrade it to `^6.4`, that'd be great. If there are new PHPUnit features you'd like to introduce to the 4.x branch, since we're now on the 6 series of PHPUnit, please feel free.

Thanks!

Potherca:
I've opened a PR (#191) to update PHPUnit on the `4.x` branch to version `^6.4`.
Once that has been merged this issue can be closed.

ramsey:
I'm closing this issue, since it's completed. Thanks!

PR #191, opened 2017-11-09
Bumps `phpunit/phpunit` development dependency to version 6.4

Potherca:
Fixes last item from #189.


ramsey:
Thanks for bumping the PHPUnit version.

I noticed that GitHub is placing an "Unverified" tag on your commits because your PGP/GPG key has expired. You might want to take a look at that.

Potherca:
Thanks for pointing out the expired GPG key. Fixed now.

PR #193, opened 2017-11-10
Add XSD schema to phpunit.xml.dist config

mhujer:
- added XSD schema, so it can be validated (and PHPStorm provides
   auto-completion)
- dropped backupGlobals option, because false is the default value


ramsey:
I was unaware of the XSD schema for phpunit.xml. Thanks!

PR #194, opened 2017-11-10
Drop paragonie/random_compat from dependencies

mhujer:
It is not necessary anymore, because the minimal required version was already set to 7.0


ramsey:
Thanks!

PR #195, opened 2017-11-10
Drop remnants of HHVM from tests

mhujer:
It was removed from Travis in c479cdd42f53c98b7f93209131c37b436503f808, so I think the extra conditions can also be dropped from the tests.


ramsey:
Thanks!

PR #197, opened 2017-12-11
[4.x] Add PHPStan analysis to the build

mhujer:
[PHPStan](https://github.com/phpstan/phpstan) is a PHP Static Analysis Tool.

I've fixed the errors it can find up to level 2 (it has 7 levels of checks), but this PR is already big enough, so I will create another one later on.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,7 @@ before_script:
 
 script:
   - ./vendor/bin/parallel-lint src tests
+  - composer run phpstan
   - ./vendor/bin/phpunit --verbose --coverage-clover build/logs/clover.xml
   - ./vendor/bin/phpcs src tests --standard=psr2 -sp
 
--- a/composer.json
+++ b/composer.json
@@ -69,7 +69,7 @@
         "phpunit": "phpunit --verbose --colors=always",
         "phpcs": "phpcs src tests --standard=psr2 -sp --colors",
         "phpstan": [
-            "@php vendor/phpstan/phpstan-shim/phpstan.phar analyse -c phpstan.neon src --level 1 --no-progress",
+            "@php vendor/phpstan/phpstan-shim/phpstan.phar analyse -c phpstan.neon src --level 2 --no-progress",
             "@php vendor/phpstan/phpstan-shim/phpstan.phar analyse -c phpstan-tests.neon tests --level 1 --no-progress"
         ],
         "test": [
--- a/composer.json
+++ b/composer.json
@@ -69,8 +69,8 @@
         "phpunit": "phpunit --verbose --colors=always",
         "phpcs": "phpcs src tests --standard=psr2 -sp --colors",
         "phpstan": [
-            "@php vendor/phpstan/phpstan-shim/phpstan.phar analyse -c phpstan.neon src --level 0 --no-progress",
-            "@php vendor/phpstan/phpstan-shim/phpstan.phar analyse -c phpstan-tests.neon tests --level 0 --no-progress"
+            "@php vendor/phpstan/phpstan-shim/phpstan.phar analyse -c phpstan.neon src --level 1 --no-progress",
+            "@php vendor/phpstan/phpstan-shim/phpstan.phar analyse -c phpstan-tests.neon tests --level 1 --no-progress"
         ],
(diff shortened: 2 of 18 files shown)


.travis.yml
```diff
@@ -25,6 +25,7 @@ before_script:
 
 script:
   - ./vendor/bin/parallel-lint src tests
+  - composer run phpstan
```
mhujer: By calling the Composer script, it runs the same configuration as you would run locally.

composer.json
```diff
@@ -37,6 +37,8 @@
         "mockery/mockery": "^1.0",
         "moontoast/math": "^1.1",
         "php-mock/php-mock-phpunit": "^2.0",
+        "phpstan/phpstan-phpunit": "0.9.2",
+        "phpstan/phpstan-shim": "0.9.1",
```
mhujer: this is the actual tool, but wrapped in a phar so it does not install any other dependencies

composer.json
```diff
@@ -67,8 +69,13 @@
         "lint": "parallel-lint src tests",
         "phpunit": "phpunit --verbose --colors=always",
         "phpcs": "phpcs src tests --standard=psr2 -sp --colors",
+        "phpstan": [
+            "@php vendor/phpstan/phpstan-shim/phpstan.phar analyse -c phpstan.neon src --level 2 --no-progress",
+            "@php vendor/phpstan/phpstan-shim/phpstan.phar analyse -c phpstan-tests.neon tests --level 2 --no-progress"
+        ],
         "test": [
             "@lint",
+            "@phpstan",
```
mhujer: it should be ran before tests, because it can discover issues that would make tests fail anyway

composer.json
```diff
@@ -67,8 +69,13 @@
         "lint": "parallel-lint src tests",
         "phpunit": "phpunit --verbose --colors=always",
         "phpcs": "phpcs src tests --standard=psr2 -sp --colors",
+        "phpstan": [
+            "@php vendor/phpstan/phpstan-shim/phpstan.phar analyse -c phpstan.neon src --level 2 --no-progress",
```
mhujer: by splitting analysis of `src` and `tests` separate config file can be provided (e.g. with different ignored erros)

src/Converter/Time/GmpTimeConverter.php
```diff
@@ -58,7 +58,7 @@ public function calculateTime($seconds, $microSeconds)
      */
     public function convertTime($timestamp)
     {
-        if (!$timestamp instanceof GMP) {
+        if (!$timestamp instanceof \GMP) {
```
mhujer: Please check if this is correct

mhujer:
Usualy
1) `/** @var Uuid $uuid */` comment is necessary anywhere, where the result of `Uuid::fromString()` is treated like the instance of  `Uuid`. (`::fromString()` returns `UuidInterface` which lacks some of the methods available in `Uuid`)

tests/Generator/SodiumRandomGeneratorTest.php
```diff
@@ -39,11 +39,11 @@ public function testFactoryUsesSodiumRandomGenerator()
...
         Uuid::setFactory($uuidFactory);
 
-        $uuid = Uuid::uuid4();
-
+        /** @var UuidFactory $actualUuidFactory */
+        $actualUuidFactory = Uuid::getFactory();
         $this->assertInstanceOf(
             SodiumRandomGenerator::class,
-            $uuid->getFactory()->getRandomGenerator()
+            $actualUuidFactory->getRandomGenerator()
```
mhujer: I'm not sure about this change, but originally it was calling the static method non-statically.

mhujer:
PHPStan works by analyzing the types in the code and in the phpdocs and then it uses it to report potentially incorrect code. (It knows, that if the method accepts some type, you cannot pass other type there.) The analysis is possible by having proper types in the code (which also helps the PHPStorm and other IDEs).

If there is anything else I can explain, you can comment directly at the specific line of the diff.

ramsey:
After considering this some more, I've decided to accept it. Thanks for all your hard work!

ramsey:
====================

PR #199, opened 2017-12-23
[4.x] Use ::class constants where possible

mhujer:
It helps with refactorings and static analysis, because you can easily tell if the class is referenced or not.

### DIFF
--- a/tests/Codec/OrderedTimeCodecTest.php
+++ b/tests/Codec/OrderedTimeCodecTest.php
@@ -29,8 +29,8 @@ class OrderedTimeCodecTest extends TestCase
     protected function setUp()
     {
         parent::setUp();
-        $this->builder = $this->getMockBuilder('Ramsey\Uuid\Builder\UuidBuilderInterface')->getMock();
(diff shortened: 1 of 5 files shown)


ramsey:
====================

| Message | Location | Level |
| ------- | -------- | ----- |
| [public] Method implementation changed | [src/Converter/Time/GmpTimeConverter.php:59](https://github.com/ramsey/uuid/blob/1d2c5ec46388b8d311e2b64de8247410cc4ef162/src/Converter/Time/GmpTimeConverter.php#L59) |  |

mabar:
@mhujer It is possible to add ::class constants everywhere, as it depends only on use statements. Autoloading is not triggered and existence of these classes is not required.

Try it yourself:
```php
use Foo\Bar;
echo Bar\Baz::class;
```

ramsey:
The reason I didn't use `::class` in pre-4.x versions of the library was because `::class` was only introduced in PHP 5.5, and this library needed to work in version 5.4. Since ramsey/uuid 4.x will no longer support PHP 5.x, this breaking change is acceptable.

Thanks!

PR #200, opened 2017-12-23
[4.x] Use specific versions of require-dev dependencies

mhujer:
It may help to prevent issues such as https://github.com/ramsey/uuid/commit/bd7f7f3135b1cd751b8be06b66a721e13f3d5b06 when a BC break is introduced in a non-semver manner.

On one hand, it means that from time to time, versions need to be updated manually. But on the other hand it allows to handle updates of dev dependencies intentionally, without colliding with other changes (e.g. fix in an older version)

_// If you don't want to merge this, just close the PR. It wasn't that much work so I created the PR directly, instead of creating an issue/discussion first._

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,18 +29,18 @@
     },
     "require-dev": {
         "ext-gmp": "*",
-        "codeception/aspect-mock": "^2.0",
-        "doctrine/annotations": "^1.2",
-        "goaop/framework": "^2.1",
-        "ircmaxell/random-lib": "^1.1",
-        "jakub-onderka/php-parallel-lint": "^0.9.0",
-        "mockery/mockery": "^1.0",
-        "moontoast/math": "^1.1",
-        "php-mock/php-mock-phpunit": "^2.0",
+        "codeception/aspect-mock": "2.1.1",
+        "doctrine/annotations": "1.4.0",
+        "goaop/framework": "2.1.2",
+        "ircmaxell/random-lib": "1.2.0",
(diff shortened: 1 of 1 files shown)


ramsey:
====================

composer.json
```diff
@@ -29,18 +29,18 @@
...
-        "codeception/aspect-mock": "^2.0",
-        "doctrine/annotations": "^1.2",
-        "goaop/framework": "^2.1",
-        "ircmaxell/random-lib": "^1.1",
-        "jakub-onderka/php-parallel-lint": "^0.9.0",
-        "mockery/mockery": "^1.0",
-        "moontoast/math": "^1.1",
-        "php-mock/php-mock-phpunit": "^2.0",
+        "codeception/aspect-mock": "2.1.1",
+        "doctrine/annotations": "1.4.0",
```
mhujer: The versions are the latest working of PHP 7.0 (there are other versions which support 7.1+ only)

GrahamCampbell:
Hmmm, not sure is the right approach. Better to establish if the upstream repos are semver or not, and ensure they follow whatever policy they say they follow.

ramsey:
I agree with @GrahamCampbell. While I hard-lock versions in project work in my company, I don't want to set explicit versions in open source projects, if I can help it. The change I made in bd7f7f3135b1cd751b8be06b66a721e13f3d5b06 for Aspect Mock was the right change, albeit an unfortunate one.

PR #205, opened 2018-01-19
added helper method for COMB UUIDs

1ma:
Resubmission of #180 with changed name helper (`timestampFirstComb()`)


1ma:
It's done. I don't really know what else to test, though.

ramsey:
After giving this a lot of thought, I've decided not to merge this pull request. In general, I like your idea of having a convenience method, but since the timestamp first COMB UUID isn't part of the UUID standard, I don't think the convenience method has a place in this library.

Thanks for contributing; keep it up!

PR #206, opened 2018-01-19
Fix for DefaultTimeGenerator

1ma:
An unrelated fix for an off-by-one error I spotted while studying your code. 16384 is actually the first 15-bit integer and [is out of spec](https://tools.ietf.org/html/rfc4122#section-4.2.2). I've changed the format to hexadecimal to emphasize where the bit width ends.

Also replaced `mt_rand` with `random_int`. Even though the library targets PHP 5.4 it already required `paragonie/random_compat` so it was already available for all versions below 7.0.

### DIFF
--- a/src/Generator/DefaultTimeGenerator.php
+++ b/src/Generator/DefaultTimeGenerator.php
@@ -79,7 +79,7 @@ class DefaultTimeGenerator implements TimeGeneratorInterface
 
         if ($clockSeq === null) {
             // Not using "stable storage"; see RFC 4122, Section 4.2.1.1
-            $clockSeq = mt_rand(0, 0x3fff);
+            $clockSeq = random_int(0, 0x3fff);
         }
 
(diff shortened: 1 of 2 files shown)


ramsey:
The `::class` constant doesn't exist in PHP 5.4, so that build is failing. Can you please update the code accordingly?

ramsey:
Strike that. It appears that test failure is from my own changes in 901c3e540d34ecc7d0761cc3c4667618f70582c1, and I didn't see it because Travis CI was having build problems that day. I'll handle fixing that. :smile:

ramsey:
Oops. I think I rebased the wrong branch or something. Let me try that one more time. Sorry. :smile:

ramsey:
Okay. That's much better. I've rebased your branch against master, since it appeared you were behind by 29 commits or so. Sorry for the noise. The build should pass on Travis CI now.

ramsey:
Thank you for contributing! 🎉

PR #209, opened 2018-01-21
Exclude docs from production

vinkla:
Added `.md` files and the docs directory to be excluded in production.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,6 +1,10 @@
-/.gitattributes   export-ignore
-/.gitignore       export-ignore
-/.travis.yml      export-ignore
-/phpunit.xml.dist export-ignore
-/tests            export-ignore
-/tools            export-ignore
+/.gitattributes     export-ignore
+/.gitignore         export-ignore
+/.travis.yml        export-ignore
+/phpunit.xml.dist   export-ignore
+/CHANGELOG.md       export-ignore
+/CODE_OF_CONDUCT.md export-ignore
+/CONTRIBUTING.md    export-ignore
+/tests              export-ignore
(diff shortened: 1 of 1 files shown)


ramsey:
Thanks for the PR.

I’ll accept exclusion of the `docs/`, but those markdown files are part of the distribution, IMO. Please see the discussion here: https://github.com/ramsey/uuid/pull/140

vinkla:
> I’ll accept exclusion of the docs/, but those markdown files are part of the distribution, IMO.

I understand your point of view. Though, these files will only be excluded in production. In most cases the `vendor` directory wont be touched in production. The markdown files will just exists without any reason.

ramsey:
The files are excluded from the zip and gz files created for each release here: https://github.com/ramsey/uuid/releases

vinkla:
They will also be excluded when running `composer update --prefer-dist` in the root of your project.

ramsey:
I know. That’s where you want them excluded. I do not want them excluded from the zip or gz distribution files.

vinkla:
> I know. That’s where you want them excluded. I do not want them excluded from the zip or gz distribution files.

Hmm, I don't think there is a way to exclude for Composer only. Shouldn't we target Composer? I'm guessing most people download the package with Composer.

ramsey:
There is not a way to target only Composer. As I’ve already said in previous discussions, to me, these files are part of the distribution in the same way the LICENSE and README.md files are part of the distribution. I do not wish to remove them from the distributed package.

vinkla:
I've pushed an update.

ramsey:
Thanks for contributing! 🎉

PR #208, opened 2018-01-20
4.x rebase fixes

mhujer:
The HHVM dropping commit got lost during the rebase.

Also some other changes got lost, so I added them here.

PR #210, opened 2018-01-28
4.x  phpstan

### LINKED ISSUE #207: Confusing phpstan errors
In the `4.x` branch, phpstan is emitting the following errors:

```
14.08s$ composer run phpstan
> phpstan analyse -c phpstan.neon src --level 2 --no-progress
 [OK] No errors                                                                 
> phpstan analyse -c phpstan-tests.neon tests --level 2 --no-progress
 ------ ------------------------------------------------------------------------- 
  Line   tests/src/Converter/Time/DegradedTimeConverterTest.php                   
 ------ ------------------------------------------------------------------------- 
         Class Ramsey\Uuid\Test\Converter\Time\DegradedTimeConverterTest was not  
         found while trying to analyse it - autoloading is probably not           
         configured properly.                                                     
 ------ ------------------------------------------------------------------------- 
 ------ ------------------------------------------------------------------------- 
  Line   tests/src/Validator/ValidatorTest.php                                    
 ------ ------------------------------------------------------------------------- 
         Class Ramsey\Uuid\Test\Validator\ValidatorTest was not found while       
         trying to analyse it - autoloading is probably not configured properly.  
 ------ ------------------------------------------------------------------------- 
 [ERROR] Found 2 errors                                                         
Script phpstan analyse -c phpstan-tests.neon tests --level 2 --no-progress handling the phpstan event returned with error code 1
```

See here: https://travis-ci.org/ramsey/uuid/builds/331056453

I don't know enough about phpstan to understand what is going on or why these errors are occurring. These test files are no longer at these locations. I moved them a while back, so I'm not sure why it's trying to do anything with them.

SignpostMarv:
* resolved #207
* switched from phpstan/phpstan-shim to phpstan/phpstan,
* adjusted the composer script to behave nicely on gitbash/travis etc.
* bumped up some of the phpstan levels, resolved issues.

a phpdoc block in moontoast/math is holding up level 6: https://github.com/ramsey/moontoast-math/blob/c2792a25df5cad4ff3d760dd37078fc5b6fccc79/src/Moontoast/Math/BigNumber.php#L633 , i.e. BigNumber is being passed in when the docs specify string|int

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -60,7 +60,7 @@
         "phpunit": "phpunit --verbose --colors=always",
         "phpcs": "phpcs src tests --standard=psr2 -sp --colors",
         "phpstan": [
-            "phpstan analyse -c phpstan.neon src --level 4 --no-progress",
+            "phpstan analyse -c phpstan.neon src --level 5 --no-progress",
             "phpstan analyse -c phpstan-tests.neon tests --level 2 --no-progress"
         ],
         "test": [
--- a/composer.json
+++ b/composer.json
@@ -60,7 +60,7 @@
         "phpunit": "phpunit --verbose --colors=always",
         "phpcs": "phpcs src tests --standard=psr2 -sp --colors",
         "phpstan": [
-            "phpstan analyse -c phpstan.neon src --level 3 --no-progress",
+            "phpstan analyse -c phpstan.neon src --level 4 --no-progress",
             "phpstan analyse -c phpstan-tests.neon tests --level 2 --no-progress"
         ],
         "test": [
--- a/composer.json
+++ b/composer.json
@@ -61,7 +61,7 @@
         "phpcs": "phpcs src tests --standard=psr2 -sp --colors",
         "phpstan": [
             "phpstan analyse -c phpstan.neon src --level 5 --no-progress",
-            "phpstan analyse -c phpstan-tests.neon tests --level 3 --no-progress"
+            "phpstan analyse -c phpstan-tests.neon tests --level 4 --no-progress"
         ],
         "test": [
             "@lint",
--- a/composer.json
+++ b/composer.json
@@ -60,7 +60,7 @@
         "phpunit": "phpunit --verbose --colors=always",
         "phpcs": "phpcs src tests --standard=psr2 -sp --colors",
         "phpstan": [
-            "phpstan analyse -c phpstan.neon src --level 7 --no-progress",
+            "phpstan analyse -c phpstan.neon src --level max --no-progress",
             "phpstan analyse -c phpstan-tests.neon tests --level 4 --no-progress"
         ],
         "test": [
--- a/composer.json
+++ b/composer.json
@@ -60,7 +60,7 @@
         "phpunit": "phpunit --verbose --colors=always",
         "phpcs": "phpcs src tests --standard=psr2 -sp --colors",
         "phpstan": [
-            "phpstan analyse -c phpstan.neon src --level 6 --no-progress",
+            "phpstan analyse -c phpstan.neon src --level 7 --no-progress",
             "phpstan analyse -c phpstan-tests.neon tests --level 4 --no-progress"
         ],
         "test": [
--- a/composer.json
+++ b/composer.json
@@ -61,7 +61,7 @@
         "phpcs": "phpcs src tests --standard=psr2 -sp --colors",
         "phpstan": [
             "phpstan analyse -c phpstan.neon src --level 5 --no-progress",
-            "phpstan analyse -c phpstan-tests.neon tests --level 2 --no-progress"
+            "phpstan analyse -c phpstan-tests.neon tests --level 3 --no-progress"
         ],
         "test": [
             "@lint",
--- a/composer.json
+++ b/composer.json
@@ -60,7 +60,7 @@
         "phpunit": "phpunit --verbose --colors=always",
         "phpcs": "phpcs src tests --standard=psr2 -sp --colors",
         "phpstan": [
-            "phpstan analyse -c phpstan.neon src --level 5 --no-progress",
+            "phpstan analyse -c phpstan.neon src --level 6 --no-progress",
             "phpstan analyse -c phpstan-tests.neon tests --level 4 --no-progress"
         ],
         "test": [
--- a/composer.json
+++ b/composer.json
@@ -60,7 +60,7 @@
         "phpunit": "phpunit --verbose --colors=always",
         "phpcs": "phpcs src tests --standard=psr2 -sp --colors",
         "phpstan": [
-            "phpstan analyse -c phpstan.neon src --level 2 --no-progress",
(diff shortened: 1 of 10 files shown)


GrahamCampbell:
> switched from phpstan/phpstan-shim to phpstan/phpstan,

why?

mhujer:
I already fixed issues from https://github.com/ramsey/uuid/issues/207 in https://github.com/ramsey/uuid/pull/208.

SignpostMarv:
@GrahamCampbell in order to use the config files defined in the phpstan package, hoping to progress towards `config.levelmax.neon` + phpstan-strict-rules

@mhujer I missed that- shall I rebase onto `mhujer:mh-4.x-rebase-fixes` ?

ramsey:
I’m still not sure what the point of all this work is. It seems like a lot of rigid overhead with little value. I’m happy to include phpstan if others are able to address problems that arise from it, but if I have to keep scratching my head at its error messages, I will likely remove it.

ramsey:
I just merged #208 into `4.x`. Should this be rebased against `4.x` before I merge it?

composer.json
```diff
@@ -60,8 +60,8 @@
         "phpunit": "phpunit --verbose --colors=always",
         "phpcs": "phpcs src tests --standard=psr2 -sp --colors",
         "phpstan": [
-            "phpstan analyse -c phpstan.neon src --level 2 --no-progress",
-            "phpstan analyse -c phpstan-tests.neon tests --level 2 --no-progress"
+            "php vendor/phpstan/phpstan/bin/phpstan analyse -c phpstan.neon src --no-progress",
+            "php vendor/phpstan/phpstan/bin/phpstan analyse -c phpstan-tests.neon tests --no-progress"
```
mhujer: Why the change from `phpstan` to the whole path? It should work the old way.

phpstan.neon
```diff
@@ -1,3 +1,5 @@
+includes:
+    - vendor/phpstan/phpstan/conf/config.level5.neon
```
mhujer: Why is this better than just passing the level from CLI parameters?

mhujer:
@ramsey this should definitely be rebased before merging it to 4.x to verify that the build still passes

composer.json
SignpostMarv: `phpstan` uses whichever version of phpstan is found under `$PATH`
`vendor/bin/phpstan` would work, except for some windows environments, i.e. git bash
`vendor/phpstan/phpstan/bin/phpstan` refers specifically to the version required by the composer.json

phpstan.neon
SignpostMarv: it might've changed recently, but there's no cli parameter for the max level, which currently aliases to `-l7`, but may change in future.

there's also the `phpstan/phpstan-strict-rules` ruleset, as well as any custom configurations that may be required in future.

composer.json
mhujer: It should always use the one installed through the Composer, see [the docs](https://getcomposer.org/doc/articles/scripts.md#writing-custom-commands).

> Note: Before executing scripts, Composer's bin-dir is temporarily pushed on top of the PATH environment variable so that binaries of dependencies are easily accessible. In this example no matter if the phpunit binary is actually in vendor/bin/phpunit or bin/phpunit it will be found and executed.

In what situation it did not work as expected?

phpstan.neon
mhujer: Yes, it's possible to [use max in CLI](https://github.com/phpstan/phpstan#rule-levels).
> You can also use `--level max` as an alias for the highest level.

composer.json
SignpostMarv: > It should always use the one installed through the Composer, see the docs.

might be an artefact of past experience/earlier development environments. checking now.

SignpostMarv: yup, I stand corrected - always nice to learn new things :)

mhujer: Great! I was afraid it is actually broken (I use similar configuration elsewhere)

phpstan.neon
SignpostMarv: well colour me corrected :P

pretty sure that importing rules still stands?

mhujer: Yep, other rules/extensions should be imported in the config.

SignpostMarv:
rebasing at the moment, dropping the phpunit configs as they seemed to have no effect on the output.

SignpostMarv:
> PHP Fatal error:  Uncaught RuntimeException: You need to provide valid cache directory for Go! AOP framework. in /home/travis/build/ramsey/uuid/vendor/goaop/framework/src/Core/AspectKernel.php:208
lolwhut ?

src/Provider/Node/FallbackNodeProvider.php
```diff
@@ -42,12 +42,12 @@ public function __construct(array $providers)
...
      * and returning the first non-empty value found
      *
-     * @return string System node ID as a hexadecimal string
+     * @return string|null System node ID as a hexadecimal string
      */
     public function getNode()
     {
         foreach ($this->nodeProviders as $provider) {
-            if ($node = $provider->getNode()) {
+            if ((bool) ($node = $provider->getNode())) {
```
mhujer: What about something like this:
```php
$node = $provider->getNode();
if ($node !== null) { 
    return $node;
}
```

When PHPStan warns you, that there is implicit boolean conversion in condition, it does not mean that it should be replaced with explicit 😄

phpstan.neon
```diff
@@ -1,7 +1,10 @@
+includes:
+    - ./vendor/phpstan/phpstan-strict-rules/rules.neon
```
mhujer: I would put this (and the necessary changes) in a separate PR.

I think it makes sense to fix the stuff PHPStan detects with the standard checks first. And then consider enabling this. (When doing both at the same time, you can't tell if the strange Travis error is related to this or not)

SignpostMarv: the changes are relatively minimal for it to pass the strict rules ?

mhujer: They probably are, so maybe it would be a wasted afford in this case.

Anyway, I just tried to build the `4.x` branch itself and it fails as well https://travis-ci.org/mhujer/uuid/jobs/341774850 - so it is probably caused by a new version of https://github.com/goaop/framework/releases

mhujer: https://github.com/ramsey/uuid/pull/213

SignpostMarv: I've been giving some thought to separate pull requests on turning the error suppression back off- tinkering with an autoload-dev shim that matches the method signatures of the missing functions & constants, but throws `BadFunctionCallException` if inadvertently called ?

SignpostMarv:
have rolled back the strict rules commits so we ca put them on a separate pr later

mhujer:
I've created a PR which fixes the build https://github.com/ramsey/uuid/pull/214

SignpostMarv:
@mhujer rebased, seems to be passing now.

ramsey:
Thank you for contributing! 🎉

Issue #187, opened 2017-10-17
Improvements to (the tests for) the SystemNodeProvider class.


Potherca:
Prompted by #185 (written by my friend and colleague @mjrider) I would like to suggest a few improvements to the `Ramsey\Uuid\Provider\Node\SystemNodeProvider` class and the test it comes with:

1. The SystemNodeProviderTest mocks (part of) the class it is testing. [This is a testing anti-pattern](http://wiki.c2.com/?SubclassToTestAntiPattern). The main reason seems to be that, instead of changing the code to be testable, the test was changed (for the worse).
I would suggest [this change](https://github.com/potherca-contrib/uuid/compare/master...07228a3b3751018399a1679fca4d3cec04e20601) to fix that. (Taking it [one step further](https://github.com/potherca-contrib/uuid/compare/2bf98b246f6ff76730cb394dae7d3a50a6c3c4b5...ddd340a9241ed0cdf83b155d60a54b7e2aa37783) might be a bit too much.)
2. <del>All tests have `@runInSeparateProcess` and `@preserveGlobalState disabled` annotations. The only reason for this seems to be the use of a static variable in the `getNode` function.
By [replacing the static variable with a class variable](https://github.com/potherca-contrib/uuid/compare/07228a3b3751018399a1679fca4d3cec04e20601...2bf98b246f6ff76730cb394dae7d3a50a6c3c4b5) this issue can easily be resolved.</del>
The static variable in the `SystemNodeProvider::getNode()` function should be class-level, not function-level

The code I link to is a suggestion, if there are any reasons I have missed, the code can be changed accordingly.

Also, as a matter of priority, I'll wait with opening a PR for this until the changes in #185 have been merged to master, as that code is in the system and these changes are more for the benefit of the tests.

---

**Edit:** After a discussion with @mjrider I understand that having `$node` be static (i.e. shared across several instances of the same class) is desirable for performance by avoiding calling the functionality more than once. I have updated my suggestion to change the level of static variable.

ramsey:
What you've described here sounds good to me. Even the "taking it one step further" with the constructor is good, IMO. If you open this as a PR, I'll accept it. Thanks!

ramsey:
@Potherca, are you still interested in working on this?

Potherca:
Yup, thanks for the reminder! I was waiting for #185 and it slipped further down my todo list over the holidays.

I've already got work for this ready on my fork, I'll take a look at it and open a PR this coming week.

[...]

PR #211, opened 2018-01-31
Improvements to the SystemNodeProviderTest class

Potherca:
As stated in [the accompanying issue](https://github.com/ramsey/uuid/issues/187), the tests for the SystemNodeProvider class suffer from the "Subclass To Test" anti-pattern by mocking the class under test.

This merge request fixes that.

Besides a minor change in the class itself, this MR focuses exclusively on the test:

- The class under test is no longer mocked, instead the native PHP functions are mocked (using the already present AspectMock).
- Documentation is added to the test class to make it's purpose more clear.
- The main test case has been expanded with a data-provider that offers more descriptive examples of the input the class under test is expected to parse.
- Two new classes (and data-providers) have been added for negative scenario's.
- The methods in the test class have been grouped so that the test functions, the mock functions and the data-provides are grouped together.

It is possible that certain test scenarios are present twice because it was not entirely clear to me what each test was trying to accomplish as the test name where somewhat lacking a descriptive quality.

The recent division of logic for a Linux specific approach has also created duplication. If so desired I could make an attempt to de-duplicate the tests.

For now, this is what I've got, feedback welcome.

### DIFF
--- a/src/Provider/Node/SystemNodeProvider.php
+++ b/src/Provider/Node/SystemNodeProvider.php
@@ -88,17 +88,16 @@ class SystemNodeProvider implements NodeProviderInterface
     {
         $mac = false;
 
-        if (strtoupper(php_uname('s')) === "LINUX") {
+        if (strtoupper(php_uname('s')) === 'LINUX') {
             $addressPaths = glob('/sys/class/net/*/address', GLOB_NOSORT);
 
             if (empty($addressPaths)) {
                 return false;
             }
 
-            $macs = array_map(
-                'file_get_contents',
-                $addressPaths
-            );
+            array_walk($addressPaths, function ($addressPath) use (&$macs) {
+                $macs[] = file_get_contents($addressPath);
+            });
 
             $macs = array_map('trim', $macs);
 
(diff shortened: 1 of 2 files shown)


Potherca:
I have no idea what's with coveralls :crying_cat_face:

ramsey:
Re: Coveralls... it's not a big deal. As long as Travis CI is passing. :smile:

src/Provider/Node/SystemNodeProvider.php
```diff
@@ -88,17 +88,16 @@ protected function getSysfs()
...
 
             if (empty($addressPaths)) {
                 return false;
             }
 
-            $macs = array_map(
-                'file_get_contents',
-                $addressPaths
-            );
+            array_walk($addressPaths, function ($addressPath) use (&$macs) {
```
ramsey: Does `array_walk()` perform better than `array_map()` or was this change made because it's easier to mock the functions and test them?

Potherca: It was done because of the mock. `array_map` performs slightly better (see [the related benchmark](https://gist.github.com/Potherca/51c7b6409f4d947c3bd7684c9cf0d96b)).

Potherca:
The Arrange / Act / Assert pattern is quite commonplace, although it is usually not called by its name.

Some useful references:

- http://wiki.c2.com/?ArrangeActAssert
- https://msdn.microsoft.com/en-us/library/hh694602.aspx#Anchor_3

PR #212, opened 2018-02-07
Add support for determining MAC address on FreeBSD systems

kalifg:
## Background
I noticed that the system node part of the UUID was always random on our FreeBSD systems.  After reviewing the code I see that `SystemNodeProvider` uses the output of `php_uname` to determine the OS it's running on.

It doesn't specifically test for FreeBSD, and the default uses a `netstat` command that isn't supported on FreeBSD.

## Solution
I have updated `SystemNodeProvider` to check for FreeBSD specifically and have included a `net stat` command that will provide output such as the following:

```
Name    Mtu Network       Address              Ipkts Ierrs Idrop    Opkts Oerrs  Coll
em0    1500 <Link#1>      08:00:27:71:a1:00    65514     0     0    42918     0     0
em1    1500 <Link#2>      08:00:27:d0:60:a0     1199     0     0      535     0     0
lo0   16384 <Link#3>      lo0                      4     0     0        4     0     0
```

This should be compatible with the current `netstat`/`ifconfig` output parser.

I have also updated `SystemNodeProviderTest` with the new system.

### DIFF
--- a/src/Provider/Node/SystemNodeProvider.php
+++ b/src/Provider/Node/SystemNodeProvider.php
@@ -70,6 +70,9 @@ class SystemNodeProvider implements NodeProviderInterface
             case 'DAR':
                 passthru('ifconfig 2>&1');
                 break;
+            case 'FRE':
+                passthru('netstat -i -f link 2>&1');
+                break;
             case 'LIN':
             default:
                 passthru('netstat -ie 2>&1');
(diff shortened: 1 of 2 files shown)


ramsey:
I took care of the conflicts.

Thank you for contributing! 🎉

kalifg:
Thank you!  I'm sorry I wasn't able to get to that sooner!

ramsey:
No worries. Thanks again!

PR #214, opened 2018-02-21
Do not use the latest goaop/framework 2.2

mhujer:
This should make the build green again.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,7 @@
         "ext-gmp": "*",
         "codeception/aspect-mock": "~2.0.0",
         "doctrine/annotations": "^1.2",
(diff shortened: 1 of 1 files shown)


ramsey:
Thanks! I think also using just `^2.1.0` would have worked, as well. 😄

tomasfejfar:
@ramsey JFYI it would not. Carret means [everything up to the next major release](https://getcomposer.org/doc/articles/versions.md#caret-version-range-). So `^2.1.0` matches `2.9.1` for example. The tilde behaves like you describe matching `<2.2.0` for `~2.1.0` constraint.

ramsey:
Sorry. I meant `~2.1.0`. :-)

Issue #142, opened 2016-10-31
Document (in composer.json) ext-ctype dependency


iansltx:
With a more minimal PHP install (e.g. Alpine packages), ctype isn't included by default, but it's required for DefaultTimeGenerator, see https://github.com/ramsey/uuid/search?utf8=%E2%9C%93&q=ctype

All well and good, but this should probably fail on composer install rather than waiting 'til runtime.

Alternately, preg_match on /^[0-9A-F]+$/i should do the trick without requiring ctype, but I have a feeling (haven't benchmarked) the ctype method is faster, and most envs will have the extension anyway.

ramsey:
Thanks for catching this, @iansltx. I'll add ext-ctype as a dependency.

ramsey:
A polyfill has been added for the Ctype functions, and I've added `ext-ctype` as a suggested extension in `composer.json` here: 735eb7e3c9a3adb3f6764677e96aa16fdd33661f

PR #223, opened 2018-04-23
Add a polyfill for ctype

BackEndTea:
This deals with #142 by adding a polyfil for the ctype functions instead of failing an install.

The only downside for end users is that their version is slightly slower if it is ran without ctype. But it does allow them to run it, instead of  not allowing the composer install.

I can understand not wanting to add a new dependency to the project, especially if it is only to support one edge-case on one function.

I can also add the following snippet to this project instead, since xdigit is the only ctype function in use.
```php
    if(!function_exists('ctype_xdigit')) {
        /**
         * Returns TRUE if every character in text is a hexadecimal 'digit', that is a decimal digit or a character from [A-Fa-f] , FALSE otherwise.
         *
         * @see http://php.net/manual/en/function.ctype-xdigit.php
         *
         * @param string|int $text
         *
         * @return bool
         */
        function ctype_xdigit($text)
        {
            $text = conver_int_to_char_for_ctype($text);
            return !empty($text) && !preg_match('/[^A-Fa-f0-9]/', $text);
        }
    }
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,8 @@
     },
     "require": {
         "php": "^5.4 || ^7.0",
-        "paragonie/random_compat": "^1.0|^2.0"
+        "paragonie/random_compat": "^1.0|^2.0",
+        "backendtea/ctype-compat": "^0.1.0"
     },
     "require-dev": {
         "moontoast/math": "^1.1",
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,7 @@
     "require": {
         "php": "^5.4 || ^7.0",
         "paragonie/random_compat": "^1.0|^2.0",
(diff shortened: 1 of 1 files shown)


ramsey:
Thanks for your contribution. Before I merge it, I notice that the version is currently `0.1.0`, and this is the only release. Do you intend to follow semantic versioning? When do you expect a 1.0.0 release?

BackEndTea:
I'd hold off on merging this since the repo will move to `symfony/polyfill` soon, https://github.com/symfony/polyfill/pull/118.

So its better to use that.

BackEndTea:
ping @ramsey I've updated this to require the symfony version.

ramsey:
Thanks!

PR #224, opened 2018-04-25
Use class-based PHPUnit namespace

peter279k:
# Changed log

As title.


ramsey:
Thank you for contributing! 🎉

PR #225, opened 2018-05-02
Don't allow UUIDs to have trailing newline

lboynton:
A simple change to disallow UUIDs with a trailing newline, which IMO should not be valid (**please correct me if I'm wrong here!**).

This change updates the regex to use the PCRE_DOLLAR_ENDONLY modifier so that it matches only if it's the end of the string.

> D (PCRE_DOLLAR_ENDONLY)
If this modifier is set, a dollar metacharacter in the pattern matches only at the end of the subject string. Without this modifier, a dollar also matches immediately before the final character if it is a newline (but not before any other newlines). This modifier is ignored if m modifier is set. There is no equivalent to this modifier in Perl.

From http://php.net/manual/en/reference.pcre.pattern.modifiers.php

### DIFF
--- a/src/Uuid.php
+++ b/src/Uuid.php
@@ -654,7 +654,7 @@ class Uuid implements UuidInterface
             return true;
         }
 
-        if (!preg_match('/' . self::VALID_PATTERN . '/', $uuid)) {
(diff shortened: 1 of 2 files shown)


ramsey:
You're correct. A newline character in a UUID should not be considered valid.

Thank you for contributing! 🎉

PR #229, opened 2018-07-02
Support random_compat v9.99.99

simonwelsh:
v9.99.99 of random_compat is a no-op library specifically for PHP7.


ramsey:
Interesting. I was unaware of this. Thanks!

Issue #222, opened 2018-04-16
Using DateTimeInterface instead of DateTime


garak:
I see that `getDateTime` method in UuidInterface (and its implementations) suggests in phpdoc comment `\DateTime` as returned type.
I was wondering if it could make sense to change it to `\DateTimeInterface`, so when one day return type will be enforced, concrete implementations could be changed to `\DateTimeImmutable` instead (that's better)

ramsey:
I agree. This would be a good change for version 4 of the library.

garak:
I guess that we can close this issue, it looks like proposed changes have been made

ramsey:
Thanks for keeping up with this. I need to get on top of things and make a big push for version 4 to get this and other things released.

PR #239, opened 2018-09-25
Change getDateTime to DateTimeInterface

Zeyckler:
This PR solves this issue https://github.com/ramsey/uuid/issues/222
Apart from that, I have changed a little bit the test to improve it, like using the const `::class` instead of the hardcoded string, removing unused returned objects and a few more improvements.


tests/UuidTest.php
```diff
@@ -8,14 +8,14 @@
 use Ramsey\Uuid\Generator\CombGenerator;
 use Ramsey\Uuid\Generator\RandomGeneratorFactory;
 use Ramsey\Uuid\Provider\Time\FixedTimeProvider;
-use Ramsey\Uuid\Provider\Time\SystemTimeProvider;
```
Zeyckler: Not used class in this test.

ramsey:
Thank you for contributing! 🎉

PR #242, opened 2018-10-01
Add UuidInterface::isValid

ninsuo:
For some reasons, there are no `isValid` method in the Uuid interface.

If that's on purpose, my apologies for this pull request 😄 

Anyway, I find very useful to check UUID validity whatever the implementation, for input validation for example. Right now I should inject the implementation which is not super nice.


ramsey:
I'm not sure it's a good idea to require all implementations to implement a static `isValid()` method. In most cases, checking validity with `Ramsey\Uuid\Uuid::isValid()` is appropriate. If you require special validation, then using your own validation mechanism is preferable.

In many cases, specifying a static method as part of a public contract (interface) seems awkward to me. The primary reason you might want to do this is so you can call `static::isValid()` from within an instance method and rely on it being present. I can see how this not being on the interface might be a problem if you start by extending `Ramsey\Uuid\Uuid`, rely on calls to `static::isValid()` internally, and then change the parent to another implementation.

I'd like to hear more about your use-case, though. Can you show an example of how you're having to inject the implementation to solve this?

ninsuo:
Hi Ramsey,

Sure you're right, `isValid` is a static method where all other getters are not... it might be a problem as well.

We are actually using 2 kind of UUIDs: 

- the classic ones, used as database ids, or as tokens, or as whatever we need simple random uuids

- uuids containing partitioning information, where by doing bitwise operations we can know that some data are or should be stored on a specific database (for example for legal reasons in Russia where russian user data should stay stored in Russia)

- uuids containing legacy information, such as historic database ids in order to reconciliate / progressively migrate data from a format to another

- etc.

The point is that we would like to create generic validators or other piece of codes where we don't have to think or know about the UUID implementation. We can handle this manually with a switch case but it removes a bit of flexibility.

ramsey:
I've decided not to accept `isValid()` as part of the interface. I do appreciate your interest in improving this library, and I hope you'll contribute new ideas and improvements in the future. Thanks!

ninsuo:
It's totally fine 👍
Cheers!

Issue #240, opened 2018-09-27
php_uname() has been disabled for security reasons


s3lcsum:
`Ramsey\Uuid\Provider\Node\SystemNodeProvider::class` uses `php_uname()` method but secure configuration of _php.ini_ do not allows calling this command.

ramsey:
### Project notes:

We had to check for a similar issue with `passthru()`. See #114 and 8d621488c4ec9c67444bc30534bc174b6bba07ee.

bkayranci:
Bug Fixes #271 

PHP_OS predefined constant was used instead of php_uname() function.

PR #244, opened 2018-10-01
Check whether php_uname is disabled

edwinkortman:
This PR solves #240 by checking if `php_uname` is disabled and returning a default value if it is.


tests/Provider/Node/SystemNodeProviderTest.php
```diff
@@ -566,10 +597,10 @@ private function assertMockFunctions(
                 $this->functionProxies[$key]->verifyNeverInvoked();
             } elseif (is_array($asserts)) {
                 foreach ($asserts as $assert) {
-                    $this->functionProxies[$key]->verifyInvokedOnce($assert);
+                    $this->functionProxies[$key]->verifyInvoked($assert);
```
edwinkortman: I am not to sure about this part. It feels hacky to either make a different flow just for `ini_get` or make it like this with just checking if it was invoked without checking how often it was invoked.

@ramsey How would you proceed?

PR #246, opened 2018-10-17
Removed some else/elseif and added some early returns

MASNathan:
Removed some `else`/`elseif` and added some early returns to make the code a little easier to read, nothing changes on the big picture


ramsey:
Thank you for contributing! 🎉

I've applied your commit to the `master` branch (3.x series) in commit https://github.com/ramsey/uuid/commit/84fe6cf755564a0374740a7ee5d7b28a29eae986. Thanks!

PR #247, opened 2018-10-17
Used __callStatic to access factory methods directly

MASNathan:
This may or may not be a good idea, let me know what you think


ramsey:
I’m not sure this is a good idea, since it exposes all the factory methods as static methods on the Uuid class.

MASNathan:
You are probably right! We can always check if the called method is one of uuid1, uuid3, uuid4, etc and throw an exception if doesn't match

ramsey:
I think it’s best to implement the methods separately on the Uuid class and let them proxy through to the factory. This way, we don’t need to check the method called or throw an exception.

MASNathan:
Gotcha

Issue #23, opened 2014-03-14
[Feature] Make uuid mockable


dimsav:
Hi and thanks for your nice package!

I tried today to write some tests to classes using `Uuid::uuid4()` but it was not possible.  The current structure of the class doesn't allow it to be mocked in unit tests since there is no way to get an instance of the class.

PR #255, opened 2019-01-07
Make Uuid final again (aka Revert 'Remove "final" keyword to allow extending')

Majkl578:
Allowing inheritance of Uuid class is a terribly bad decision and encourages bad design:
- There is *46* (!) public methods in Uuid class, these implicitly become public API of any descendant.
- People abuse inheritance to avoid writing few lines of code without thinking of consequences.
- Composition should be encouraged and even enforced, it shouldn't be _up to the consumer (opposed to _" I'll agree with you that I prefer composition over inheritance, but I prefer choice even more so."_ by @stanlemon in #25).
- The argumentation in #23 is wrong, Uuid is not supposed to be mocked, it's a value object. Also there is an interface (now).

More on this topic: https://ocramius.github.io/blog/when-to-declare-classes-final/

As this is a BC break for anyone performing heretical practices, targets next major.

This reverts commit c7daee6c6a582110a1d60bb482a9644202d3c7a2 / PR #25.

_(The build is failing now, will address that once discussed.)_


ostrolucky:
What actual, existing problems are you solving with this massive BC break? You are fixing _potential_ problems which are yet to happen over the years, unlike reverse with final class which was already issue in past. Please link references which confirm this class being non-final negatively affected class design decisions because of BC concerns, or people complaining changes broke their subclasses.

ramsey:
It’s a BC break for those who might be extending the class to override functionality. My understanding is that there are some who like to inject their own bits into the UUID at certain places. I think the design changes I’m proposing in #256 provide a better path for this, but it does break BC for anyone currently extending `Uuid`.

OskarStark:
Regardless of being a BC break here or not, it can be handled and a new major could be released including `final`, so existing code will still work.

If I understand @Majkl578 correctly this PR is just about doing it or not (not about "don't do it, it's a BC break").
This is a ValueObject with an interface, which allow mocking very easily if needed (this is often an argument against making a class final), so this should definitely be final.

For me this is more a decision about the **future** and **not the past,** closing the API and knowing how its gonna be used is a big plus and will bring more flexibility for further development.

I am 👍 

//cc @greg0ire

Ocramius:
> Should protected fields and methods become private?

Can be done safely, plus can apply all the optimisations you want now :+1:

unkind:
> and UuidInterface already provides an excellent way to implement alternate UUID variations.

By the way, this interface is better be removed as well.

OskarStark:
Can we proceed here with a new major?

ramsey:
This most definitely will be included in a new major version. I have some structural changes I intend to make and have not had time to devote to this, hence the delay.

ramsey:
I just attempted to cherry-pick this commit into the new `4.x` branch and realized that it won't work in its current state because `DegradedUuid` extends `Uuid`. 🤦‍♂

PR #262, opened 2019-07-14
add php 7.4snapshot

andreybolonin:


ramsey:
That’s a lot of deprecation warnings and errors for `ReflectionType::__toString()`! 😮 

It looks like most of these are coming from phpstan, so maybe they have a newer version that works well on 7.4?

A lot of the ramsey/uuid tests also appear to use it, so we’ll need to find a work-around and update all the broken tests.

carusogabriel:
@ramsey Hey man, how are you doing?

I'd like to work on `ramsey/uuid` forward compatibility on behat of github.com/usabilla. Does this work need to be done on the `master` branch, or can we make it in the `3.x`?

ramsey:
@carusogabriel Sorry for taking so long to respond. I would love to have help on forward compatibility! I've just ported many things that were in the `4.x` branch (old `master` branch) into the `3.x` branch, and I made the `3.x` branch the new `master`. If you could make your changes off the `master` branch (3.x), that would be great. I'm going to continue development off of that branch for now.

ramsey:
@andreybolonin, thank you for contributing! 🎉

I've added 7.4 as an "allowed failure" to the `master` branch. It currently isn't available on Travis CI, and when they do make it available, there are too many errors right now, as a result of a change to the Reflection interface that goaop/parser-reflection doesn't follow. I'll need to follow up with them to see when a fix is expected.

PR #263, opened 2019-07-27
Function api created

mnavarrocarter:
The function api is a set of simple namespaced autoloaded functions to create uuids proxying the Uuid class static factory methods.

They provide easy and clean ways of just creating string represented uuids instead of the class.

It think it would be a useful adition to the library to have some functions that just return strings, mainly because of simplicity and code readability.

The functions are: 
- Ramsey\Uuid\v1
- Ramsey\Uuid\v3
- Ramsey\Uuid\v4
- Ramsey\Uuid\v5

They take the same arguments as their simils in the Uuid class static methods, and also have exactly the same doc block except for the return type.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -50,7 +50,10 @@
     "autoload": {
         "psr-4": {
             "Ramsey\\Uuid\\": "src/"
-        }
+        },
+        "files": [
+            "src/functions.php"
+        ]
     },
     "autoload-dev": {
         "psr-4": {
--- a/src/functions.php
+++ b/src/functions.php
@@ -1,0 +1,59 @@
+<?php
+
+namespace Ramsey\Uuid;
+
+/**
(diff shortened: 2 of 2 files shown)


ramsey:
I've applied this commit to the `master` branch (3.x series): https://github.com/ramsey/uuid/commit/2ec28dd736ac58606e110bda2c4dbc84672c2c3e

I made a few changes to make this work in the 3.x series:

- https://github.com/ramsey/uuid/commit/be88f95a50ed738578fab4b489a578eaa0aa81e8
- https://github.com/ramsey/uuid/commit/35cf50d7b408fdfc2bc78508a0dafe1778b90354

Thank you for contributing! 🎉

mnavarrocarter:
@ramsey Just saw this. Thanks for merging! :)

PR #270, opened 2019-10-26
PHP_OS predefined constant was used instead of php_uname() function.

### LINKED ISSUE #240: php_uname() has been disabled for security reasons

bkayranci:
Bug Fixes #240 

PHP_OS predefined constant was used instead of php_uname() function.


ramsey:
Thank you for contributing! 🎉

PR #271, opened 2019-10-26
PHP_OS predefined constant was used instead of php_uname() function.

### LINKED ISSUE #240: php_uname() has been disabled for security reasons

bkayranci:
Bug Fixes #240 

PHP_OS predefined constant was used instead of php_uname() function.

### DIFF
--- a/src/Provider/Node/SystemNodeProvider.php
+++ b/src/Provider/Node/SystemNodeProvider.php
@@ -67,7 +67,7 @@ class SystemNodeProvider implements NodeProviderInterface
         }
 
         ob_start();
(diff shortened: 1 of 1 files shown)


ramsey:
I transitioned the `3.x` branch to `master`, and I merged #270 into `master`. I will be using a different strategy to rebase the `4.x` branch, so what's in `master` now will all make its way into `4.x`.

PR #273, opened 2019-11-07
Backport type casting Uuid to string

### LINKED ISSUE #272: UuidInterface should declare method __toString() in version 3.x
The `Uuid` class already implements this method as a thin wrapper around `toString()`, however, since it's not declared in the interface, code-quality tools will trigger an error if you type-cast an instance of Uuid to string.

For example, the following code will work

```php
<?php

echo (string) \Ramsey\Uuid\Uuid::uuid4();
echo PHP_EOL;
```

But phpstan will trigger the following error

```
 ------ -------------------------------------------------- 
  Line   src/uuid_repro.php                                
 ------ -------------------------------------------------- 
  5      Cannot cast Ramsey\Uuid\UuidInterface to string.  
 ------ -------------------------------------------------- 
```

I prefer type casting over calling toString() since, in my opinion, makes the code more readable. Locally, I simply added `public function __toString();` to the interface and the error is no longer reported, but I'm not sure how this will interact with the rest of the library (I'm only using it to generate a uuid4).

Pnoexz:
Fixes #272

### DIFF
--- a/src/UuidInterface.php
+++ b/src/UuidInterface.php
@@ -275,4 +275,11 @@ interface UuidInterface extends JsonSerializable, Serializable
      * @return string
      */
     public function toString();
+
+    /**
+     * Converts this UUID into a string representation.
+     *
+     * @return string
+     */
+    public function __toString();
 }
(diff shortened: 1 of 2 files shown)


ramsey:
Thanks for your contribution! Since this would be a significant BC break for the 3.x series, I can't accept it, so this will be available in 4.0.0. I'll release 4.0.0 before the end of the year.

PR #279, opened 2019-12-18
illegal access of properties at Uuuid.php

gabrielpires:
## Description
The `unserialize` function at the Uuid.php creates an instance of Uuid from a serialized object.
After this, with this new instance, mutates its own properties: `codec`, `converters`, and `fields` based on that instance. 
The problem is: codec, converters, and fields are protected properties in some versions and even private in most recent versions. This is a violation according to PHP documentation.
Calling `unserialize`  at this point, will cause a problem.

To fix the problem, I created 2 new public functions: `getCodec` and `getConverters`, and change the calls inside of the `unserialize` function.

## Motivation and context
This change needs to be done for the `unserialize` function does not throw an exception.
I didn't find a specific issue related, so I decide to make the fix and open a Pull Request.

## How has this been tested?
The new version (this one that I'm submitting for PR) was tested inside a project that I'm working on (my employer's project). The exception was not thrown anymore.
Also ran there  `composer run-script test` the make sure nothing was changed.

## Types of changes

## Checklist:
- [X] I have read the **[CONTRIBUTING.md](https://github.com/ramsey/uuid/blob/master/.github/CONTRIBUTING.md)** document.
- [X] All new and existing tests passed.
- [X] I have run `composer run-script test` locally, and there were no failures or errors.

### DIFF
--- a/src/Uuid.php
+++ b/src/Uuid.php
@@ -6,10 +6,10 @@
  * file that was distributed with this source code.
  *
  * @copyright Copyright (c) Ben Ramsey <ben@benramsey.com>
- * @license   http://opensource.org/licenses/MIT MIT
- * @link      https://benramsey.com/projects/ramsey-uuid/ Documentation
- * @link      https://packagist.org/packages/ramsey/uuid Packagist
- * @link      https://github.com/ramsey/uuid GitHub
+ * @license http://opensource.org/licenses/MIT MIT
+ * @link https://benramsey.com/projects/ramsey-uuid/ Documentation
+ * @link https://packagist.org/packages/ramsey/uuid Packagist
+ * @link https://github.com/ramsey/uuid GitHub
  */
 
 namespace Ramsey\Uuid;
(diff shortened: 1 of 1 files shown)


gabrielpires:
@ramsey sorry about it, my IDE just added those. Revert the changes and now only the essential code was pushed.

ramsey:
Thanks for doing that. The code looks good.

Can you show a snippet of code that reproduces the exception that you're seeing? I'm using the following script to serialize and unserialize UUIDs both with `$codec`, `$converter`, and `$fields` set to `protected` and with them set to `private`, and I've run this on PHP 5.4 through 7.4, and I don't see any errors or exceptions.

``` php
<?php

use Ramsey\Uuid\Uuid;

$uuid = Uuid::uuid4();

$serialized = serialize($uuid);

$uuid2 = unserialize($serialized);

echo "\n";
echo "UUID pre-serialization  : " . $uuid->toString();
echo "\n";
echo "UUID post-serialization : " . $uuid2->toString();
echo "\n\n";

if ($uuid->equals($uuid2)) {
    echo "The two UUIDs are equal.\n";
} else {
    echo "The two UUIDs are NOT equal.\n";
}
```

ramsey:
Access to the private/protected properties is allowed in this context because they are objects of the same type.

For reference, see http://docs.php.net/language.oop5.visibility#language.oop5.visibility-other-objects

> Objects of the same type will have access to each others private and protected members even though they are not the same instances. This is because the implementation specific details are already known when inside those objects.

ramsey:
Bump. :smile:

Can you paste some code to reproduce the exception/error you are getting? Thanks!

ramsey:
Following up on this... Please show how you're getting an exception/error. I'd like to get a better understanding of the issue this PR fixes.

ramsey:
Checking in again. If I don't hear back soon, I'll close this PR, but feel free to open a new one if the error is still occurring and you can present an example that shows how it's happening. Thanks!

ramsey:
Closing due to lack of response. Feel free to open a new issue, if you continue to experience problems. Thanks!

PR #281, opened 2019-12-27
Minimal UUID immutability invariants at type-level

Ocramius:
## Description

This patch adds `@psalm-immutable` to `UuidInterface`, and marks the named constructors of `Uuid` as `@psalm-pure` too.

This is very similar to https://github.com/moneyphp/money/pull/576 in spirit and implementation

## Motivation and context

UUIDs are identifiers that cross serialization and address space boundaries, and should be treated as immutable values.

## How has this been tested?

A new `static-analysis` (not exported) directory has been created: it includes fixtures that are verifying the transitivity of the added `@psalm-pure` and `@psalm-immutable` type invariant markers.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

Please note that this will cause type changes for downstreams implementors (custom implementations) of `UuidInterface` that do use `vimeo/psalm`: if their code does not follow `@psalm-immutable` invariants, they will get a build failure. No runtime behavior changes apply, so no new warnings/exceptions.

## Checklist:

- [x] ~~My change requires a change to the documentation.~~ (inapplicable)
- [x] ~~I have updated the documentation accordingly.~~ (inapplicable)

### DIFF
--- a/src/Uuid.php
+++ b/src/Uuid.php
@@ -649,6 +649,9 @@ class Uuid implements UuidInterface
      *
      * @return UuidInterface A UuidInterface instance created from a binary
      *     string representation
+     *
+     * @psalm-pure note: changing the internal factory is an edge case not covered by purity invariants,
+     *             but under constant factory setups, this method operates in functionally pure manners
      */
     public static function fromBytes(string $bytes): UuidInterface
     {
@@ -662,6 +665,9 @@ class Uuid implements UuidInterface
      *
      * @return UuidInterface A UuidInterface instance created from a hexadecimal
      *     string representation
+     *
+     * @psalm-pure note: changing the internal factory is an edge case not covered by purity invariants,
+     *             but under constant factory setups, this method operates in functionally pure manners
      */
     public static function fromString(string $uuid): UuidInterface
     {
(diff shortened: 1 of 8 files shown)


src/Uuid.php
```diff
@@ -649,6 +649,9 @@ public static function setFactory(UuidFactoryInterface $factory): void
      *
      * @return UuidInterface A UuidInterface instance created from a binary
      *     string representation
+     *
+     * @psalm-pure note: changing the internal factory is an edge case not covered by purity invariants,
```
Ocramius: Important to know: `@psalm-`-prefixed annotations are ignored by current (known) annotation scanner libraries

src/UuidInterface.php
```diff
@@ -129,7 +131,7 @@ public function getDateTime(): DateTimeInterface;
     /**
      * Returns the 128-bit integer value of the UUID as a string
      *
-     * @return mixed
+     * @return string
```
Ocramius: This happened to be fixed during static analysis runs.

@psalm prevents usage of `mixed` (good) when configured with `totallyTyped="true"`: this was probably left un-documented to avoid excluding the `SlevomatCodingStandard.TypeHints.ReturnTypeHint.MissingNativeTypeHint` rule from CS standards configuration (see change in `phpcs.xml.dist` in this patch)

ramsey:
Thanks, @Ocramius! This looks great. 🎉

Ocramius:
Nice, thanks for the extremely speedy reaction! \o/

PR #287, opened 2020-01-06
Use brick math package instead of moontoast

RSpeekenbrink:
Use the ``brick/math`` package for BigNumbers instead of the abandoned ``moontoast/math`` package.

## Description

Replaced all ``Moontoast/Math/BigNumber`` usages with the ``Brick/Math/BigInteger`` variant.

## Motivation and context

This change has been made since the currently used package will no longer be maintained according to the message on packagist.

## How has this been tested?

Tested this by running the provided PHPUnit tests.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

I'm not sure in how far this is a breaking change. I tried to keep everything compatible like I kept the BigNumber name in the Converters instead of renaming them to i.e. BigIntegerTimeConverter although that would take more sense. As far as the package goes it does not support the getValue() function on BigIntegers but just passes the value in the __toString() function.

It does require a replacement of a package so does that qualify it for being breaking?

## Checklist:

- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,12 +7,6 @@ services:
 jobs:
   fast_finish: true
   include:
-    - php: 5.4
-      dist: trusty
-    - php: 5.5
-      dist: trusty
-    - php: 5.6
-    - php: 7.0
     - php: 7.1
     - php: 7.1
       arch: s390x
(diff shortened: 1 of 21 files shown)


RSpeekenbrink:
``brick/math`` requires PHP 7.1 at least. So this actually is breaking since it requires PHP ^7.1.

ramsey:
Thanks for your willingness to jump in and make this change. I’ve already made the change to the 4.x branch, though, but I haven’t pushed the changes to the repo yet. Sorry for the confusion. 🙂

PR #289, opened 2020-01-10
Added custom getValue() method for custom APIs

shehi:
Added custom getValue() method for custom APIs

## Description
There are no changes here really, just added an alias method.

## Motivation and context
In some APIs, id fields are objects, which when serialized, try to get ID values via a ```getValue()``` method. In such cases, since ```UUID::getValue()``` doesn't exist, we get stuck. Adding this method is easier compared to adding a workaround to such APIs. E.g. in medical standard FHIR/R4, id's are resources which have ```value``` field in them, with ```getValue()``` method attached. Unfortunately changing standards like this is literally impossible.

## How has this been tested?
It's an alias to an existing method, doesn't really need any testing.

## Types of changes

## Checklist:

### DIFF
--- a/src/Uuid.php
+++ b/src/Uuid.php
@@ -228,6 +228,17 @@ class Uuid implements UuidInterface
     {
         return $this->toString();
     }
+    
+    /**
+     * Converts this UUID object to a string. Needed for custom APIs which require
+     * an object here with 'value' field (e.g. FHIRResource::getId() in FHIR R4 medical standard).
+     *
+     * @return string
+     */
+    public function getValue()
+    {
+        return $this->toString();
+    }
 
     /**
      * Re-constructs the object from its serialized form.
--- a/tests/UuidTest.php
+++ b/tests/UuidTest.php
@@ -1897,6 +1897,16 @@ class UuidTest extends TestCase
         $this->assertEquals('"' . $uuid->toString() . '"', json_encode($uuid));
     }
 
+    /**
+     * @covers Ramsey\Uuid\Uuid::getValue
+     */
+    public function testGetValue()
+    {
+        $uuid = Uuid::uuid1();
+
+        $this->assertEquals('"' . $uuid->getValue() . '"', json_encode($uuid));
+    }
+
     public function testSerialize()
     {
         $uuid = Uuid::uuid4();


shehi:
@ramsey , do I need to write a test for this. Addition is just an alias to an existing method.

shehi:
Ok added a test. Didn't want to decrease your coverage :)

ramsey:
I understand that the requirements you referenced require a `value` property / `getValue()` method, and I think this can be accomplished in your system using something like composition. For example:

``` php
class MyIdentifier
{
    private $uuid;

    public function __construct(\Ramsey\Uuid\UuidInterface $uuid)
    {
        $this->uuid = $uuid;
    }

    public function getValue(): string
    {
        return $this->id->toString();
    }
}
```

My preference is not to include an alias, since this is something that will ultimately need to be added to the interface so that all UUID objects have the method, and to require everyone implementing the interface to implement both `toString()` and `getValue()`, which do the same thing, seems like an additional burden.

Can you provide more information about how you're using `Ramsey\Uuid\Uuid` in your application? Are you using an ORM that automatically returns database IDs as `Uuid` objects, for example?

shehi:
@ramsey , unfortunately I can't use it that way. Thing is, as I mentioned, its a medical standard and library in question is this: https://github.com/dcarbone/php-fhir-generated .

This library unfortunately can't be altered. According to FHIR/R4 standard, all resources eventually boil down to (https://github.com/dcarbone/php-fhir-generated/blob/master/src/DCarbone/PHPFHIRGenerated/R4/FHIRResource.php)[```FHIRResource``` class].

ramsey:
It sounds like you're using the library in a way that's not supported. If a PHP 7 version of the library is ever released that uses return typehints, you'll run into a number of problems, since `getId()` is supposed to return a `FHIRId` instance instead of a `UuidInterface`.

I think the proper approach here is to set up your Doctrine ODM to map identifiers to a custom class that extends `FHIRId`. When Doctrine pulls the UUID out of the database and populates this custom class, you can convert the value to a `Ramsey\Uuid\Uuid` and store it in a property on that class. Then, you can implement the `getValue()` method to return the string UUID.

shehi:
It is not really relevant, regarding PHP-7, because I know what version I am using. Currently this implementation is the only way we can go, because FHIR spec isn't really Doctrine friendly. And creating separate implementation of it isn't cost effective. This isn't really our current concern.

As to ```FHIRId``` implementation, it still isn't usable: Doctrine documents/entities need to have their identifiers directly in them, not in their children/references, or in children of children, or in references of references. So this is the only way to proceed. My ODM mapping works flawlessly - I can convert data from Db to PHP and vice versa with zero problems. The problem at hand is solely DRY related: duplicated the content of 3 ```jsonSerialize()``` methods in an inheritance chain VS adding one single line in your library. If you can't or don't prefer to do that, then please let me know. I guess I will have to fork your lib at that point, because a UUID library against big FHIR library isn't really a choice to consider. No offense! :)

I already implemented your lib as an ODM bundle in [johanderuijter/uuid-doctrine-odm](https://github.com/johanderuijter/uuid-doctrine-odm) package (you only have ORM one as a bundle). So again, I don't have any problems regarding the usage of your package as ODM type. It works flawlessly.

Thanks.

ramsey:
I'd prefer not to add this for a single use-case, especially since I believe you can accomplish what you're trying to do in multiple ways that don't require modifying the `Ramsey\Uuid\Uuid` class directly.

One thing you could do is create your own UUID class and swap out the UUID builder used by the `UuidFactory`. If you take this approach, you probably won't need to make any changes to the ODM/ORM, as long as you do this early in your application's bootstrap.

It would look something like this:

``` php
<?php

use Ramsey\Uuid\Builder\UuidBuilderInterface;
use Ramsey\Uuid\Codec\CodecInterface;
use Ramsey\Uuid\Codec\StringCodec;
use Ramsey\Uuid\Converter\NumberConverterInterface;
use Ramsey\Uuid\FeatureSet;
use Ramsey\Uuid\Uuid;
use Ramsey\Uuid\UuidFactory;

class MyUuid extends Uuid
{
    public function getValue()
    {
        return $this->toString();
    }
}

class MyUuidBuilder implements UuidBuilderInterface
{
    private $converter;

    public function __construct(NumberConverterInterface $converter)
    {
        $this->converter = $converter;
    }

    public function build(CodecInterface $codec, array $fields)
    {
        return new MyUuid($fields, $this->converter, $codec);
    }
}

$factory = new UuidFactory();

$builder = new MyUuidBuilder($factory->getNumberConverter());
$codec = new StringCodec($builder);

$factory->setUuidBuilder($builder);
$factory->setCodec($codec);

// This is where the magic happens. You're telling the class to use your
// customized factory instead of the default one.
Uuid::setFactory($factory);

$uuidA = Uuid::fromString('3980191c-340f-11ea-850d-2e728ce88125');
$uuidB = Uuid::uuid4();

echo $uuidA->getValue() . "\n";
echo $uuidB->getValue() . "\n";
```

shehi:
As mentioned before, that won't work unless I want to get rid of johanderuijter/uuid-doctrine-odm package - I am using UUID through it. Anyway, thanks @ramsey .

shehi:
On additional note, I apologize as this PR seemed to be premature: I eventually had to rewrite all ```jsonSerialize()```rs so Uuid ```getValue()``` part got fixed now.

PR #292, opened 2020-01-29
Refining type of producing a `string` that can be declared as `non-empty-string`

Ocramius:
## Motivation and context

Overall, we know that a UUID can never be an empty string, hex, or numeric string. Even the zero-case is `'0'` in hex/int stringy form.

This patch formalizes that, allowing consumers of `UuidInterface` to rely on the fact that the produced `string` will never be empty.

## How has this been tested?

Additional static-analysis tests were introduced: they just check variance.

## Types of changes

## Checklist:
- [x] ~~My change requires a change to the documentation.~~
- [x] ~~I have updated the documentation accordingly.~~

### DIFF
--- a/tests/static-analysis/UuidIsNeverEmpty.php
+++ b/tests/static-analysis/UuidIsNeverEmpty.php
@@ -1,0 +1,48 @@
+<?php
+
+/**
+ * This file is part of the ramsey/uuid library
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ *
+ * @copyright Copyright (c) Ben Ramsey <ben@benramsey.com>
+ * @license http://opensource.org/licenses/MIT MIT
+ */
+
+declare(strict_types=1);
+
+namespace Ramsey\Uuid\StaticAnalysis;
+
+use Ramsey\Uuid\UuidInterface;
+
+/**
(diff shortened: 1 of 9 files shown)


tests/static-analysis/UuidIsNeverEmpty.php
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    /** @psalm-return non-empty-string */
+    public function integerIsNeverEmpty(UuidInterface $uuid): string
+    {
+        return $uuid->getInteger();
+    }
+
+    /** @psalm-return non-empty-string */
+    public function stringIsNeverEmpty(UuidInterface $uuid): string
```
Ocramius: @muglug I could not verify `__toString()` in this patch: does psalm maybe override the type of `__toString()` with `string` regardless of explicit docblock additions?

Ocramius: https://github.com/vimeo/psalm/issues/2712

ramsey:
One of the jobs timed out on the build, causing the failure. I've restarted that job.

Ocramius:
@ramsey this one should be good to go as-is. The missing test for `#__toString()` is not worth waiting for a `vimeo/psalm` release, IMO :+1:

Ocramius:
Thanks! :dancer:

PR #293, opened 2020-01-29
Import used functions and constants via `use` statement to disallow ambiguity for compiler/static-analysis

Ocramius:
As documented in https://wiki.php.net/rfc/use_global_elements, the engine (by default) does a local namespace
lookup, then falls back to global namespace when first calling a global function referenced in namespaced code,
unless that function is referenced via fully qualified name (FQN).

By using the FQN, the actual symbol can be looked up at compile-time, both by the PHP engine and by static analysis
tooling, allowing for compiler (in particular) optimizations to replace known hot-path functions with specialized opcodes.

Sadly, no actual benchmark at hand: the improvements can be minimal or massive, depending on where this library is
used (tight loops being most relevant).

## How has this been tested?

Sadly, no benchmark is provided: improvements are purely theoretical, but since the change is mostly tracked by phpcs, this makes for trivial code changes and maintenance.

## Types of changes

## Checklist:

- [x] ~~My change requires a change to the documentation.~~
- [x] ~~I have updated the documentation accordingly.~~
- [x] ~~I have added tests to cover my changes.~~ (inapplicable? Maybe needs a benchmark design overall?)


ramsey:
This will cause a bunch of tests to fail, since we're replacing some of these functions with mocks. While this may not be a good practice, it's the only way I can control the tests with known values. I'm open to ideas for how to rework these tests to avoid mocking the global functions.

Ocramius:
If specific mocking is used in non-hot-paths, we can exclude those files from CS rules: for most pure functions, this should be OK and not affecting tests.

Let's see which tests fail, and then we can decide whether to ignore the rule in CS, and do a one-off import this time.

Ocramius:
Removed any imports for functions that were replaced via `AspectMock::func()`. Also disabled the rule that forces any function to be imported (can be re-enabled if/when there is an exclusion map). Let's see how the build goes...

Ocramius:
No dice: was hoping to run things through travis, since I don't have the same stack locally. Will need more work.

ramsey:
I've provided a Vagrantfile that you can use locally: https://github.com/ramsey/uuid/tree/master/resources/vagrant/linux

Or, you can use one of the arm32 Docker images I have here: https://hub.docker.com/repository/docker/benramsey/ramsey-uuid

Ocramius:
Yeah, apparently can't run ARMv7 @ 32bit on my host system, and virtualbox is loooooooooong gone :| 

I'll keep slamming travis for the time being

ramsey:
Oops! Yeah. That's an issue. Oddly enough, I can run it on macOS, but on Linux, the host CPU needs to be ARM. On Travis CI, I have to select arm64 as the host OS for those builds.

I just pushed amd64 images that have the same environment on them:

* https://hub.docker.com/layers/benramsey/ramsey-uuid/php-7.4-amd64/images/sha256-3b8dbd256b96d6babe135694c30ec5d66cb929ceb2201926a349cb2d00b79054
* https://hub.docker.com/layers/benramsey/ramsey-uuid/php-7.4-amd64-without-xdebug/images/sha256-726e6437e4a252f8cec5caa5a065e1c0549743c00d92ec367fe4e6e1f7af1298

Ocramius:
habemus viridi! (besides the usual stuck tasks)

ramsey:
Nice! I thought that was a Harry Potter reference, so I had to look it up. Latin for "I have green?"

Ocramius:
Yup :)

"Habemus Papam" is something they used to say for when the new pope was chosen, once spending a countless days deciding which old geezer would be the next patriarchy big boss.

I like making fun of it :P

ramsey:
So, this is good to go?

Ocramius:
Yass

ramsey:
Thanks! Looks great!

Issue #306, opened 2020-04-07
Guid::uuid4() doesn't generate a Guid


prolic:
`echo get_class(Ramsey\Uuid\Guid\Guid::uuid4());`

will return

`Ramsey\Uuid\Rfc4122\UuidV4`.

I didn't expect this behavior at all.

prolic:
Also consider this:

```
$f = new Ramsey\Uuid\UuidFactory(new Ramsey\Uuid\FeatureSet(true));

var_dump($f->uuid4());
```

will return:

`PHP Fatal error:  Uncaught Ramsey\Uuid\Exception\InvalidArgumentException: The byte string received does not contain a valid version in vendor/ramsey/uuid/src/Guid/Fields.php:83`

ramsey:
Are these two different issues, or do you consider them related?

As for the 1st issue, please see #301. The static methods are not part of the interface and should not be used from the subclasses. This is a bug. I’m sorry for the confusion.

prolic:
As I am trying to create Guids in both cases, those issues are related to
me, maybe from a technical perspective they aren't for you.

I can create separate issues if you want to.

ramsey:
The first issue is an oversight on my part. You should not have access to `Guid::uuid4()`. This is a bug. Only the `Ramsey\Uuid\Uuid` class should have the static methods. I will roll out a release soon that disables these static methods on child classes (they are only child classes for the sake of BC and easy transition from version 3 to 4; they will no longer be child classes in version 5).

I’ll look into the second issue. At a glance, this looks like it should work, so I think you’ve uncovered another bug here.

Thanks!

prolic:
> $f = new Ramsey\Uuid\UuidFactory(new Ramsey\Uuid\FeatureSet(true));
> 
> var_dump($f->uuid4());

This is solved in master branch already, not working on 4.0.1 though.

ramsey:
The master branch is the same as 4.0.1 right now. The only difference is white space. How can this work in master and not in 4.0.1?

prolic:
Can't tell, that's what I tried. Don't trust, verify.

enumag:
@ramsey It's not fixed. The bug just doesn't always appear, it's random. I posted more details in the separate issue: https://github.com/ramsey/uuid/issues/314#issuecomment-734339547

PR #307, opened 2020-04-07
Allow static guid calls

prolic:
Allow static calls to Guid::uuid4() and others

see https://github.com/ramsey/uuid/issues/306

Maybe that's not the very best approach, but it should do the trick.


prolic:
If you don't like this and want to block static calls on Guid instead, maybe you should also consider removing (or deprecating for now) those static methods from Uuid class as well. To me this would be kind of weird to have Uuid::uuid4() allowed but Guid::uuid4() not. If you have to go through the UuidFactory to create those, this should be the case for all then.

ramsey:
I'm giving more thought to this. At first, I wanted to restrict calls to these static methods on the inheriting classes, but I'm giving this a second thought and will let you know soon. Thanks!

PR #309, opened 2020-04-08
Added Uuid:FromHexadecimal() and enhanced Uuid:FromString()

pluk77:
As requested.

I opted to re-use FromString in FromHexadecimal as that would catch any short Heexadecimal numbers.


src/Codec/StringCodec.php
```diff
@@ -118,6 +118,10 @@ protected function getBytes(string $encodedUuid): string
             $encodedUuid
         );
 
+        if (strpos($parsedUuid, '0x') === 0) {
+            $parsedUuid = substr($parsedUuid, 2);
+        }
```
ramsey, line 123: I think this belongs in `UuidFactory::fromHexadecimal()`, before passing the value to the codec for decoding.

src/UuidFactoryInterface.php
```diff
@@ -179,4 +179,16 @@ public function fromDateTime(
...
+     * Creates a UUID from a Hexadecimal object
+     *
+     * @param Hexadecimal $hex Hexadecimal object representing a hexadecimal number
+     *
+     * @return UuidInterface A UuidInterface instance created from the Hexadecimal 
+     * object representing a hexadecimal number
+     *
+     * @psalm-pure
+     */
+    public function fromHexadecimal(Hexadecimal $hex): UuidInterface;
```
ramsey, line 193: I'm unable to add this to `UuidFactoryInterface` until version 5, since it's a major breaking change to anyone who is implementing this interface.

ramsey (changes requested):

src/Codec/StringCodec.php
pluk77: Reason for adding it here is to allow the hex string to be used via this method. See the new test cases.

src/UuidFactoryInterface.php
pluk77: Makes sense. Will remove it from the interface.

src/Codec/StringCodec.php
ramsey: I think this should throw an exception, since it's not a valid string UUID:

``` php
Uuid::fromString('0x1EA78DEB37CE625E8F1A025041000001')
```

The main problem here is what to do if someone configures the library to use a different codec. Then, neither `fromString()` nor `fromHexadecimal()` will work as expected.

pluk77: Makes perfect sense. Will update accordingly.

pluk77:
Is there anything else you need from me before you can consider this for inclusion in  your bundle?

ramsey:
I've been swamped lately, so I haven't had a chance to get back to this. Are you able to resolve the failures? If so, I can go ahead and merge this. If not, it will need to wait until I have some time to fix them.

Thanks!

pluk77 (commented):
Applied the requested changes

pluk77:
I don't think I can resolve these failures as some originate form the fact the the new method can not be added to the uuidFactoryInterface for BC purposes. I will wait patiently for you to have some time.

pluk77:
@ramsey How is your workload at the moment? Would you be able to look at the failed integration test as it seems like it can only be resolved by adding the function to the factory interface and thereby breaking backwards compatibility.

ramsey (approved):

src/Uuid.php
```diff
@@ -514,6 +514,22 @@ public static function fromDateTime(
...
+     *
+     * @return UuidInterface A UuidInterface instance created from the Hexadecimal
+     * object representing a hexadecimal number
+     *
+     * @psalm-pure note: changing the internal factory is an edge case not covered by purity invariants,
+     *             but under constant factory setups, this method operates in functionally pure manners
+     */
+    public static function fromHexadecimal(Hexadecimal $hex): UuidInterface
+    {
+        return self::getFactory()->fromHexadecimal($hex);
```
ramsey: This is failing static analysis because `getFactory()` returns a `UuidFactoryInterface`, which doesn't have this method on it. This can cause problems if using a factory that doesn't have this method.

```suggestion
        $factory = self::getFactory();
        
        if (\method_exists($factory, 'fromHexadecimal')) {
            return self::getFactory()->fromHexadecimal($hex);
        }

        throw new \BadMethodCallException('The method fromHexadecimal() does not exist on the provided factory');
```

ramsey: I'm going to commit this suggestion, though it might still fail coding standards checks because of the use of the root namespace without `use` statements. I'll fix those, though.

PR #333, opened 2020-08-23
Use assertSame instead of assertEquals

Bilge:
All occurrences of `assertEquals` have been replaced with `assertSame`.

## Description
*Assert equals* coerces types if they are different whereas *assert same* never performs type coercion. The differences are similar to the differences between loose comparison (`==`) and strict comparison (`===`). 

## Motivation and context
For the same reason that strict quality (`===`) is always preferred over loose quality (`==`), *assertSame* should always be preferred over *assertEquals* because *assertEquals* can mask bugs and provide false positives. Having made the changes, it seems there were no such false positives, but we can only be sure since making this change.

In some cases, *assertEquals* was inspecting private members of objects to perform equality comparisons, such as in the following example.

https://github.com/ramsey/uuid/blob/cd4032040a750077205918c86049aa0f43d22947/tests/Rfc4122/FieldsTest.php#L216-L224

These have been made compatible with *assertSame* by calling the public interface methods on the objects instead.

https://github.com/ramsey/uuid/blob/7947413a76e5e5fa427198a3d14d73fbfd61de0a/tests/Guid/FieldsTest.php#L202

## How has this been tested?
Ran tests locally.

## Types of changes

## Checklist:

### DIFF
--- a/tests/Codec/GuidStringCodecTest.php
+++ b/tests/Codec/GuidStringCodecTest.php
@@ -106,7 +106,7 @@ class GuidStringCodecTest extends TestCase
 
         $codec = new GuidStringCodec($this->builder);
         $result = $codec->decode($string);
-        $this->assertEquals($this->uuid, $result);
+        $this->assertSame($this->uuid, $result);
     }
 
     public function testDecodeBytesReturnsUuid(): void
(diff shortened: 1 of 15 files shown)


ramsey:
Thank you for contributing! 🎉

PR #334, opened 2020-08-23
Fixed static methods invoked as instance methods

Bilge:
All methods declared as static are now invoked statically (`::`) instead of dynamically (`->`).

## Description
Static methods should be invoked statically. PHP is quirky in that this behaviour is not enforced, but since this project currently uses a mixture, these have been normalized in the interests of consistency.

## Motivation and context
Besides consistency, invalid invocation style presents a slew of warnings in my editor as pictured. This goes some way towards combating that.

image

Here's some examples of mixed invocation styles in the current version.

### Static invocation
https://github.com/ramsey/uuid/blob/d103e07a2cb745414992ab3b4a0cd9fa5d783296/tests/UuidTest.php#L1059-L1071

### Dynamic invocation
https://github.com/ramsey/uuid/blob/d103e07a2cb745414992ab3b4a0cd9fa5d783296/tests/UuidTest.php#L1030-L1031

## How has this been tested?
Ran tests locally.

## Types of changes

## Checklist:

### DIFF
--- a/docs/testing.rst
+++ b/docs/testing.rst
@@ -41,7 +41,7 @@ create a regular UuidV1 instance and pass it.
         $uuid = Uuid::uuid1();
         $myObj = new MyClass();
 
-        $this->assertIsString($myObj->tellTime($uuid));
+        self::assertIsString($myObj->tellTime($uuid));
     }
 
 This might satisfy our testing needs if we only want to assert that the method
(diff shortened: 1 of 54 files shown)


ramsey:
I'm confused about this change. All PHPUnit documentation shows examples using these as instance methods rather than static methods. Has PHPUnit changed it's practices?

ramsey:
https://phpunit.readthedocs.io/en/9.3/assertions.html#static-vs-non-static-usage-of-assertion-methods

I'm inclined to leave these as instance method calls because the test method is invoked on a test object.

Bilge:
In terms of purism, only static invocation of static methods is correct. In terms of pragmatism, neither is more correct than the other because both work in identical fashion. However, regarding the PHPUnit documentation, it is my opinion that their stance claiming *neither is correct* is simply more comfortable than admitting they misguided everyone for decades into doing the wrong thing.

ramsey:
I'm going to stick with personal preference and convention on this. 🙂

PR #336, opened 2020-08-23
Added factory method for UUIDv4 with monotonic prefix

Bilge:
Added a factory method for UUIDv4 with monotonic prefix (COMB codec). 

## Description
[The documentation](https://uuid.ramsey.dev/en/latest/customize/timestamp-first-comb-codec.html) describes a method of augmenting the standard UUIDv4 generation with an algorithm that substitutes a monotonic prefix. This PR wraps up that code in a static factory method that's easier to use.

## Motivation and context
A weakness of the current implementation is that it requires the user to fiddle with a number of internal structures, including codecs and random number generators, which should be considered an implementation detail that users don't want to be burdened with. In particular, this becomes problematic if the code breaks in future due to upgrades and changes to implementation details because users cannot be expected to know how to fix this. By moving the invocation to a static factory method, it is the library's responsibility to make sure this functionality continues to work via a simpler construction mechanism, which is much easier for users to consume and reduces the chance that their code breaks in future.

## How has this been tested?
Ran tests locally.

## Types of changes

## Checklist:

- [x] My change requires a change to the documentation.


src/Generator/CombGenerator.php
```diff
@@ -107,7 +107,7 @@ public function generate(int $length): string
 
         return (string) hex2bin(
             str_pad(
-                bin2hex((string) $hash),
```
Bilge, line 110: Redundant cast: it is impossible for `$hash` to be anything other than a string at this location due to initialization to a string and string type hints on invoked methods.

ramsey:
I rejected a similar PR (#205) a few years ago. I'll have to give this some thought. I want to encourage folks to start using v6 (and v7, when implemented) UUIDs, instead of the COMB UUIDs.

Bilge:
Why is v6 preferred over COMB?

Bilge:
If I may share my thoughts... this comes down to one of two things: either COMB is part of this library now and for the future or it is not and will be deprecated and scheduled for removal. If it's going to be removed then this PR is redundant but if it's going to remain available in future versions then this factory method is necessary. This is in spite of whether you prefer v6 over COMB because making COMB hard to use is not the correct way to discourage people from using it. If v6 is strictly better than COMB then this should be documented with an explanation as to why, but that is a separate issue.

On the topic of v6 vs COMB, it seems to me that for users who just want a random UUID with monotonic prefix, COMB is a better fit than v6, because v6 requires a lot more code to achieve the same thing. That is, it requires the user to implement randomizing the node and the clock sequence manually, which is hinted at in the documentation but doesn't come with any examples for clock sequences.

tl;dr: if COMB has a future as part of this library, it should be easy to use for those whom wish to use it.

ramsey:
Right. v6 is not technically the same. However, v7 (not yet implemented) is effectively the same as a timestamp-first COMB. v6 and v7 are also non-standard, but I'm hoping to make them standard.

Bilge:
OK, but v7 doesn't exist yet, so for the time being doesn't it warrant a static factory method for COMB?

ramsey:
Thank you for your work on this.

Since this library now supports v6 and v7 UUIDs. I will deprecate both the ordered-time and timestamp-first COMB codecs, since they are no longer necessary.

PR #339, opened 2020-09-05
remove paragonie/random_compat reference

mantis:
Remove reference to paragonie/random_compat - this library only applies before php 7 (https://github.com/paragonie/random_compat/commit/35469243ebbb9be5086cff57538cac6da2c36852)  and base uuid has been bumped to require php 7.2

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,6 @@
         "goaop/framework": "^2",
         "mockery/mockery": "^1.3",
         "moontoast/math": "^1.1",
-        "paragonie/random-lib": "^2",
         "php-mock/php-mock-mockery": "^1.3",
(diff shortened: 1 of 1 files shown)


ramsey:
Thank you for your contribution. However, [paragonie/random-lib](https://github.com/paragonie/RandomLib) is different from [paragonie/random_compat](https://github.com/paragonie/random_compat). This library provides an [adapter for RandomLib](https://github.com/ramsey/uuid/blob/0d7b8c2b7a99a5bf69e96b84ead75c74290a5fb2/src/Generator/RandomLibAdapter.php), and removing this causes the tests for that adapter to fail.

PR #340, opened 2020-09-08
Add interface for UUID-related exceptions

shrikeh:
## Description
As the exception thrown during the creating of an instance of a UUID requires intimate knowledge of the codec/builder relationship, I had a delve, and thought the simplest answer would be to give them an interface. This allows catching a `UuidExceptionInterface $e` which will keep the necessary knowledge lighter in the enclosing class.

## How has this been tested?

I ran the existing test suite, after seeing you are catching on specific exceptions. Because the use of the interface to catch is optional, it didn't look like I needed to refactor there, as the tests are after the more specific detail of the exception anyway.

## Types of changes

## Checklist:

- [x] I have run `composer run-script test` locally, and there were no failures or errors. (Well there was one but not related, phan complained about `Method Ramsey\Uuid\Converter\Number\GenericNumberConverter::fromHex() should return string&numeric but returns string.` So it's not my fault ;-)


shrikeh:
@ramsey Oh mighty god of uuids, what do you think?

ramsey:
Well, I'm not a fan of being acknowledged as a deity. I'm more of a demigod. 😉

I like the feature. I need to decide whether this goes into the next major version or the current 4.x series. Since it introduces an interface and doesn't change one, I *think* it still fits semver to go into the next minor.

shrikeh:
Morning saviour.

While it’s a bit early for more Monty Python quotes, my coffee-withdrawn brain agrees with you that it fits the minor version semver definition, because it’s a backwards compatible feature.

ramsey:
@shrikeh I've released this in v4.2.0. Thanks!

PR #350, opened 2020-10-29
Allow phpunit 9

remicollet:
phpunit 9 is the only version with official PHP 8 support

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,7 @@
         "phpstan/phpstan": "^0.12",
         "phpstan/phpstan-mockery": "^0.12",
         "phpstan/phpstan-phpunit": "^0.12",
-        "phpunit/phpunit": "^8.5",
+        "phpunit/phpunit": "^8.5 || ^9",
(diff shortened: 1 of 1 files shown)


ramsey:
Thanks. I tried getting this to work last night but had to back out, since there were so many problems with goap/framework and PHPUnit 9. I think some of the dependencies that get updated cause goap/framework to have problems. I'm in the process of looking into replacing goap/framework with something else.

ramsey:
This is the primary error that goap/framework is causing:

```
PHP Fatal error:  A function with return type must return a value in /home/runner/work/uuid/uuid/src/Uuid.php on line 374
```

It appears to rewrite the source code, so even though that's pointing to `src/Uuid.php`, the file with the fatal error is actually at `build/cache/goap/src/Uuid.php`.

ramsey:
I merged this in separately in 2f92bdf34e45a71626f51e69a85c038555bf0c8e. Thanks!

PR #351, opened 2020-10-29
use expectError instead of expectException

remicollet:
Using PHPUnit 9

```
1) Ramsey\Uuid\Test\Generator\CombGeneratorTest::testGenerateWithOddNumberOverTimestampBytesCausesError
Support for using expectException() with PHPUnit\Framework\Error\Error is deprecated and will be removed in PHPUnit 10. Use expectError() instead.

```

`expectError` already exists in PHPUnit 8

### DIFF
--- a/tests/Generator/CombGeneratorTest.php
+++ b/tests/Generator/CombGeneratorTest.php
@@ -127,7 +127,7 @@ class CombGeneratorTest extends TestCase
 
         $generator = new CombGenerator($randomGenerator, $converter);
 
-        $this->expectException(PHPUnitError::class);
+        $this->expectError();
         $generator->generate(7);
     }
 }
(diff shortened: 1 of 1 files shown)


remicollet:
Notice, 2 other warnings need to be fixed (don't find the easy way for now)


```
2) Ramsey\Uuid\Test\Generator\RandomLibAdapterTest::testGenerateUsesGenerator
Trying to configure method "generate" which cannot be configured because it does not exist, has not been specified, is final, or is static

3) Ramsey\Uuid\Test\Generator\RandomLibAdapterTest::testGenerateReturnsString
Trying to configure method "generate" which cannot be configured because it does not exist, has not been specified, is final, or is static

```

ramsey:
Thank you for contributing! 🎉

PR #359, opened 2021-01-07
Allow install of paragonie/random_compat 9.99.100

ausi:
## Description
Allow installation of `paragonie/random_compat` version 9.99.100 to avoid downgrades to version 2.x when upgrading a project to PHP 8.

## Motivation and context
`paragonie/random_compat` 9.99.99 is not compatible with PHP 8, but 9.99.100 is, see [`v9.99.99...v9.99.100`#diff-d2ab9925ca](https://github.com/paragonie/random_compat/compare/v9.99.99...v9.99.100#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34)

## How has this been tested?
`composer require 'ramsey/uuid:^3.8'` with PHP 8 installs `paragonie/random_compat` in the wrong version (`2.0.19` instead of `9.99.99`)

## Types of changes

## Checklist:

- [ ] I have run `composer run-script test` locally, and there were no failures or errors.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,7 @@
     "require": {
         "php": "^5.4 | ^7 | ^8",
         "ext-json": "*",
-        "paragonie/random_compat": "^1 | ^2 | 9.99.99",
+        "paragonie/random_compat": "^1 | ^2 | ^9.99.99",
         "symfony/polyfill-ctype": "^1.8"
     },
     "require-dev": {


composer.json
```diff
@@ -23,7 +23,7 @@
     "require": {
         "php": "^5.4 | ^7 | ^8",
         "ext-json": "*",
-        "paragonie/random_compat": "^1 | ^2 | 9.99.99",
+        "paragonie/random_compat": "^1 | ^2 | ^9.99.99",
```
dshoreman, line 26: It's probably better to require `>= 1` here rather than explicit major versions, in line with packages like defuse/php-encryption (which requires `>= 2`). v3 or v10 are unlikely, but never say never ;)

```suggestion
        "paragonie/random_compat": ">= 1",
```

dshoreman (commented):
+1, composer flagged this up for me via travis when I added PHP 8.

```
Installing dependencies from lock file (including require-dev)

Verifying lock file contents can be installed on current platform.

Your lock file does not contain a compatible set of packages. Please run composer update.

  Problem 2

    - paragonie/random_compat is locked to version v9.99.99 and an update of this package was not requested.

    - paragonie/random_compat v9.99.99 requires php ^7 -> your php version (8.0.1) does not satisfy that requirement.

  ...

  Problem 6

    - paragonie/random_compat v9.99.99 requires php ^7 -> your php version (8.0.1) does not satisfy that requirement.

    - ramsey/uuid 3.9.3 requires paragonie/random_compat ^1 | ^2 | 9.99.99 -> satisfiable by paragonie/random_compat[v9.99.99].

    - ramsey/uuid is locked to version 3.9.3 and an update of this package was not requested.
```

ramsey:
Thank you for contributing! 🎉

composer.json
ramsey: Using semver with `>= 1`, the problem is that v3 or v10 are allowed to break BC with v1 and v2, which could cause more problems. In the highly unlikely event there is a v3 or higher in paragonie/random_compat, we'll update this library. 😄

dshoreman: Sounds good :)

dshoreman:
@ramsey Thanks for merging this. After a quick update with `ramsey/uuid:3.x-dev`, Composer is happy again!

Since in my case it's actually a sub-dependency, would a 3.9.4 release be possible so it can be picked up by Composer automatically?

ramsey:
I am so sorry it took me so long to tag a 3.9.4 release. I thought I had already done so but then came across this during some recent clean up and found that it was still outstanding. I'm so sorry! 😳

PR #383, opened 2021-08-09
Fix soft BC break with `Uuid::fromString()` signature

cs278:
The `$uuid` parameter on `Uuid::fromString()` [was changed from `string` to `non-empty-string`](https://github.com/ramsey/uuid/commit/c005f69d6e5c16a807093d693ed1dbdf730b3c67#diff-898c45cfd882f04a4877c82e14ce28181f8ea6062654772bfb9fc474014bd6a0R439) this breaks BC for people using static analysis in their downstream projects.

## Description

I've reverted the change to the docblock and used an assertion to teach static analysis tools when the `$uuid` variable is a non empty string instead.

## Motivation and context

Whilst this isn't a language level backwards compatibility break, it does break when consumers are using static analysis tools on their own code base because they'd need to add checks to test if the input is non empty before calling `Uuid::fromString()`.

## How has this been tested?

I've added a unit test covering the call to `Uuid::fromString('')`.

## Types of changes

## PR checklist

### DIFF
--- a/src/Uuid.php
+++ b/src/Uuid.php
@@ -24,6 +24,7 @@ use Ramsey\Uuid\Rfc4122\FieldsInterface as Rfc4122FieldsInterface;
 use Ramsey\Uuid\Type\Hexadecimal;
 use Ramsey\Uuid\Type\Integer as IntegerObject;
 
+use function assert;
 use function bin2hex;
 use function preg_match;
 use function str_replace;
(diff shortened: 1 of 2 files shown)


Jamesking56 (approved):

ramsey:
Thanks for catching this, and thank you for contributing! 🎉

PR #388, opened 2021-09-05
Add support for PHP 8.1

GrahamCampbell:
This PR gets this library working on PHP 8.1. The new PHP 7.4 style serialisation is now mandatory, from PHP 8.1 onwards. I've checked that the two incompatible serialisation formats are in fact compatible. The old style can be deserialized just fine: https://3v4l.org/XfUS3. This means upgrading is easy and no risk. On PHP 7.2 and 7.3, there are no changes. On PHP 7.4+, the old style gets deserialized properly, and going forward, the new style will be used for serialisation.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -66,7 +66,9 @@ jobs:
           - "7.3"
           - "7.4"
           - "8.0"
-          - "8.1"
+        include:
+          - php-version: "8.1"
+            composer-options: "--ignore-platform-req=php"
 
     steps:
       - name: "Checkout repository"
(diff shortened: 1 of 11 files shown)


GrahamCampbell:
Hmm, tried to get it to upload coverage from two different runs, hoping it would merge them. Seems not. We need PHP less than 7.4 and PHP at least 7.4 to get 100% coverage on these changes, together.

GrahamCampbell:
Friendly ping, @ramsey. cc @nunomaduro

ramsey:
I'll try to take a look this week. 🙂

ramsey (approved):

ramsey:
Thank you for contributing! 🎉

PR #390, opened 2021-09-18
Fix phpunit deprecation warning

abenerd:
PHPUnit currently spits out warning about the usage of a deprecated function, I refactored the code to use the suggested replacement.
## Description

When running the test suite phpunit spits out the following warning
```bash
assertRegExp() is deprecated and will be removed in PHPUnit 10. Refactor your code to use assertMatchesRegularExpression() instead.
```
I replaced the usage of `assertRegExp` with `assertMatchesRegularExpression`.
## Motivation and context

It removes the usage of currently deprecated code.
## How has this been tested?

The testcase that has been updated still passes.
## Types of changes

## PR checklist

### DIFF
--- a/tests/Provider/Node/SystemNodeProviderTest.php
+++ b/tests/Provider/Node/SystemNodeProviderTest.php
@@ -91,7 +91,7 @@ class SystemNodeProviderTest extends TestCase
             'Node should be a hexadecimal string of 12 characters. Actual node: %s (length: %s)',
             [$node->toString(), strlen($node->toString()),]
         );
-        $this->assertRegExp('/^[A-Fa-f0-9]{12}$/', $node->toString(), $message);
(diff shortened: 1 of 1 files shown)


ramsey (approved):

ramsey:
Thanks for contributing!

Unfortunately, this fails on PHP 7.2 :-/

```
Error: Call to undefined method Ramsey\Uuid\Test\Provider\Node\SystemNodeProviderTest::assertMatchesRegularExpression()
```

abenerd:
We could check whether the method exists before calling it and using the old deprecated one as a fallback. But I think for now it would be easier to just wait until support for PHP 7.2 is dropped before changing this. PHPUnit 10 is also going to be a while still, so the tests should work still until we use that.

ramsey:
Sounds good. Thanks!

Ocramius:
This is fixed in #405

ramsey:
Thanks, @Ocramius !

PR #395, opened 2021-12-24
Use dependabot to suggest updates to all actions

Lewiscowles1986:
## Description

RE: https://twitter.com/Ocramius/status/1474179909157273601
from
https://twitter.com/ramsey/status/1474178806923595778

## Motivation and context

I'd like to see dependabot offer a suggestion for upgrading the action like Ocramius suggested; it might have broader impact than just solving the upgrade.

## How has this been tested?

It's from GitHub docs, and modifies their example frequency to match your composer frequency. I Figure if you don't want to be bombarded daily or weekly with composer; I should probably modify their example to be the same frequency rather than in-tandem advocating to spend more of your time on maintenance.

## Types of changes

- [X] New feature (non-breaking change which adds functionality)

## PR checklist

I don't believe the PR checklist applies as this is declarative config yaml; but feel free to correct me; Merry Christmas and a Happy New Year

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,3 +5,7 @@ updates:
     schedule:
       interval: "monthly"
     versioning-strategy: "increase-if-necessary"
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "monthly"


ramsey:
Many, many thanks!

I'll go ahead and merge this and see if Dependabot will send us a PR to update the package. (I hope it's something that will happen shortly and not a month from now. 😄 )

(As an aside: I noticed those PHPUnit segmentation faults yesterday. They're only happening on the code coverage job; I'm not sure why.)

PR #401, opened 2022-01-28
LazyUuidFromString for valid UUIDs in uppercase

hubipe:
This pull request tries to improve the Uuid::fromString static class by creating LazyUuidFromString even for UUID string which are not lowercase.

## Description
This pull request solves discrepancy of returning LazyUuidFromString for UUIDs, which are not in lower case. LazyUuidFromString::VALID_REGEX is case sensitive, so only UUID strings in lower case are accepted. In the same time, the function strtolower() is called in constructor parameter, but the if statement already accepted only strings in lowercase. Thus calling the strtolower on string in lowercase is redundant.
I updated the method so it lowers the string first and then tries to match the regexp. If it is valid uuid string according to the regexp, then instance of LazyUuidFromString is returned. Otherwise the functionality is not affected.

## Motivation and context
When reading data from MS SQL, their GUID strings are returned in uppercase. Calling Uuid::fromString creates for example UuidV4 instance, but most of the time, the LazyUuidFromString object would be sufficient (there is no need for parsing the UUID) (specially when UuidInterface is used for methods typehinting).

## Types of changes

## PR checklist

### DIFF
--- a/src/Uuid.php
+++ b/src/Uuid.php
@@ -476,10 +476,11 @@ class Uuid implements UuidInterface
      */
     public static function fromString(string $uuid): UuidInterface
     {
+        $uuid = strtolower($uuid);
         if (! self::$factoryReplaced && preg_match(LazyUuidFromString::VALID_REGEX, $uuid) === 1) {
             assert($uuid !== '');
 
-            return new LazyUuidFromString(strtolower($uuid));
+            return new LazyUuidFromString($uuid);
         }
 
         return self::getFactory()->fromString($uuid);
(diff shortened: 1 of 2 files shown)


ramsey:
Hi, @hubipe! Thanks for the PR. Can you provide a test for your change?

hubipe:
Hello @ramsey, I provided two unit tests. Are they sufficient?

ramsey (approved):
Thanks!

ramsey:
Tests look good. The static analysis error is not a result of your changes, so no need to worry about it.

Thank you for contributing! 🎉

PR #405, opened 2022-03-22
Bump to PHP 8.0, remove polyfills, require ext-ctype, lock dependencies, improve dependency upgrade automation

Ocramius:
## Description

 * targets **4.3.0**
 * drop all polyfills
 * upgrade the requirement to PHP 8.0
 * require `ext-ctype` to be installed
 * lock dependencies to their latest version
 * improve @dependabot automation to reduce maintenance burden on @ramsey's side

## Motivation and context

The main change here is the dropping of `symfony/polyfill*` dependencies: these dependencies are supposed to be installed by consumers, and not by libraries. If libraries start requiring the polyfills, they:

 1. become a dependency in the dependency graph, increasing ecosystem complexity
 2. shadow actual functionality that might need to be enabled on target systems (i.e. "works best with ext-ctype")
 3. introduce a dependency to an ecosystem that is highly unstable (bad track record of `symfony/*` packages and their supported version ranges)

While supporting these changes, I noticed that a local `composer install` led me to a project that is in a non-workable state:

 * psalm report failed
 * phpunit reported warnings

In order to improve on that, I also added `composer.lock` to the diff: this project uses @dependabot on a monthly basis, so I think it is a valuable addition to run on a locked dependency state, upgraded monthly. I've added automation to make sure that @dependabot can merge any green transition, therefore reducing the burden on @ramsey having to verify only any patches that transition the build state to red.

## Types of changes

- [ ] ~~Bug fix (non-breaking change which fixes an issue)~~
- [ ] ~~Breaking change (fix or feature that would cause existing functionality to change)~~

## PR checklist

- [ ] ~~My change requires a change to the documentation.~~ not needed
- [ ] ~~I have updated the documentation accordingly.~~ not needed
- [ ] ~~I have added tests to cover my changes.~~ not needed

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,12 +9,11 @@
     ],
     "license": "MIT",
     "require": {
-        "php": "^7.2 || ^8.0",
+        "php": "^8.0",
+        "ext-ctype": "*",
         "ext-json": "*",
         "brick/math": "^0.8 || ^0.9",
-        "ramsey/collection": "^1.0",
-        "symfony/polyfill-ctype": "^1.8",
-        "symfony/polyfill-php80": "^1.14"
+        "ramsey/collection": "^1.0"
     },
     "replace": {
         "rhumsaa/uuid": "self.version"
--- a/.gitattributes
+++ b/.gitattributes
@@ -8,6 +8,7 @@
 /build/                     export-ignore
 /captainhook.json           export-ignore
 /CHANGELOG.md               export-ignore
+/composer.lock              export-ignore
 /codecov.yml                export-ignore
 /CODE_OF_CONDUCT.md         export-ignore
 /CONTRIBUTING.md            export-ignore
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,6 +1,8 @@
 version: 2
 updates:
   - package-ecosystem: "composer"
+    allow:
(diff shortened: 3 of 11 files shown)


composer.json
```diff
@@ -9,12 +9,11 @@
...
     "require": {
-        "php": "^7.2 || ^8.0",
+        "php": "^8.0",
+        "ext-ctype": "*",
         "ext-json": "*",
         "brick/math": "^0.8 || ^0.9",
-        "ramsey/collection": "^1.0",
-        "symfony/polyfill-ctype": "^1.8",
-        "symfony/polyfill-php80": "^1.14"
+        "ramsey/collection": "^1.0"
```
Ocramius, line 16: @ramsey how do you feel about also dropping this (in a future PR)? As far as I can see, the collection is only used in `BuilderCollection`, but `list<UuidBuilderInterface>` would most likely suffice for the use-case of this library, in my opinion :thinking: 

Obviously to be calibrated and discussed elsewhere, but it's something I noticed while working here.

Ocramius: I'll think of a mitigation path in a new PR 👍

Ocramius: Resolving here meanwhile, thanks for the input!

ramsey (approved):

PR #406, opened 2022-03-22
Allow `iterable<T>` where `ramsey/collection` would be used, deprecate collection classes

Ocramius:
This patch is based on https://github.com/ramsey/uuid/pull/405 (merge that first)
    
As discussed in https://github.com/ramsey/uuid/pull/405#discussion_r832095457, the `ramsey/collection`
dependency can be phased out.
    
Doing so via direct removal is not feasible without a clear BC break, so this change only deprecates:
    
 * `Ramsey\Uuid\Builder\BuilderCollection`
 * `Ramsey\Uuid\Provider\Node\NodeProviderCollection`
    
These classes will then be removed in `5.0.0`, along with the `ramsey/collection` dependency.

### DIFF
--- a/src/Builder/BuilderCollection.php
+++ b/src/Builder/BuilderCollection.php
@@ -27,6 +27,11 @@ use Traversable;
 /**
  * A collection of UuidBuilderInterface objects
  *
+ * @deprecated this class has been deprecated, and will be removed in 5.0.0. The use-case for this class comes from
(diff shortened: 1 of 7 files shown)


Ocramius:
Note: coverage obviously decreased, as I phased out usage of the deprecated classes :)

ramsey (approved):

PR #410, opened 2022-04-28
Feature/assert non empty string

gsteel:
## Description

Adds a psalm assertion to `Uuid::isValid()` to assert the input is also a non-empty-string when validation passes.

## Motivation and context

Improves static analysis for consumers

## How has this been tested?

Added code to `tests/static-analysis/ValidUuidIsNonEmpty.php` that fails psalm checks without the assertion.

I don't know how to get PHPStan to accept the test returns a non-empty-string type - I'm unfamiliar with PHPStan and it doesn't seem to support Psalm's assert-if-true annotation??

## Types of changes

## PR checklist

### DIFF
--- a/phpstan-tests.neon
+++ b/phpstan-tests.neon
@@ -26,3 +26,7 @@ parameters:
                 - ./tests/Guid/*Test.php
                 - ./tests/Nonstandard/*Test.php
                 - ./tests/Rfc4122/*Test.php
+        -
+            message: "#^Method Ramsey\\\\Uuid\\\\.+ should return non-empty-string but returns string\\.$#"
+            paths:
+                - ./tests/static-analysis/ValidUuidIsNonEmpty.php
--- a/src/Uuid.php
+++ b/src/Uuid.php
@@ -537,6 +537,8 @@ class Uuid implements UuidInterface
      *
      * @psalm-pure note: changing the internal factory is an edge case not covered by purity invariants,
      *             but under constant factory setups, this method operates in functionally pure manners
+     *
+     * @psalm-assert-if-true non-empty-string $uuid
      */
     public static function isValid(string $uuid): bool
     {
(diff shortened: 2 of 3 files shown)


ramsey:
I'd like to accept and merge this, but it looks like the test is failing:

> Method Ramsey\Uuid\StaticAnalysis\ValidUuidIsNonEmpty::validUuidsAreNotEmpty() should return non-empty-string but returns string.

gsteel:
Having searched high and low… I don't think PHPStan can be made to understand the conditional assertion:

https://phpstan.org/r/6cfede4f-99c0-4d6e-b656-25d55ecd4998
https://psalm.dev/r/e7d19b38fd

I've added a suppression to PHPStan test config to silence the error, but please feel free to close if that's not acceptable!

ramsey:
> I've added a suppression to PHPStan test config to silence the error

@ondrejmirtes, do you know a way to accomplish this in PHPStan?

ondrejmirtes:
Currently only with a custom type-specifying extension. But the support for `@psalm-assert` / `@phpstan-assert` is in the works: https://github.com/phpstan/phpstan-src/pull/1317

ramsey:
@Ocramius, can you take a look at this and let me know what you think. I trust your judgment on the impact of these annotations more than I trust my own. 😄

Ocramius:
This is problematic if a `non-empty-string` is already given as input, since the assertion will be detected as redundant 🤔

gsteel:
> This is problematic if a `non-empty-string` is already given as input, since the assertion will be detected as redundant

Psalm and PHPStan don't seem to have a problem with it: [799515f](https://github.com/ramsey/uuid/pull/410/commits/799515f3087c437532c5884ad82ba1ea4b05a29f)

Ocramius:
Oh, if redundancy is not detected, then I wouldn't mind :+1: 

Still some funky behavior (observe that `never` :D ) @ https://psalm.dev/r/bf6564bf87

Redundant asserts also not being detected: https://psalm.dev/r/1ebf3d2ce6

gsteel:
The `never` is weird… It persists even with a body of `return $input !== '';`

PR #418, opened 2022-05-13
Fix upgrading manual link in README.md

SilverFire:


README.md
```diff
@@ -38,7 +38,7 @@ composer require ramsey/uuid
 
 See the documentation for a thorough upgrade guide:
 
-* [Upgrading ramsey/uuid Version 3 to 4](https://uuid.ramsey.dev/en/latest/upgrading/3-to-4.html)
+* [Upgrading ramsey/uuid Version 3 to 4](https://uuid.ramsey.dev/en/4.3.0/upgrading/3-to-4.html)
```
ramsey: This was my mistake. Read the Docs links `latest` to docs built from the default branch, which I changed to `5.x`, and I didn't want folks to think docs generated from `5.x` were stable, so I disabled the `latest` docs in Read the Docs, but I forgot to turn on the `stable` docs, which builds from the most recent release tag.

I've now turned on `stable`, so this link can be updated to use it.

```suggestion
* [Upgrading ramsey/uuid Version 3 to 4](https://uuid.ramsey.dev/en/stable/upgrading/3-to-4.html)
```

ramsey (approved):
Thanks for catching this!

ramsey:
Thank you for contributing! 🎉

PR #424, opened 2022-06-20
Allow brick/math 0.10

remicollet:


remicollet:
Looks like something wrong in new brick/math

Reported as https://github.com/brick/math/issues/66

remicollet:
CI failure also happens on 8.1 with brick/math 0.9

BenMorel:
Not sure why it fails with brick/math 0.9.x as `5.x` passes. Let me know if I can help!

remicollet:
> Not sure why it fails with brick/math 0.9.x as `5.x` passes. Let me know if I can help!

Explained in #425 ;)   (so not related to brick/math version)

alexander-schranz:
I think this would make sense on the `4.x` branch? I also think the old `moontoast/math` could be removed I could not find any usage of it in the code base while tried creating the same PR: https://github.com/alexander-schranz/uuid/commit/8659bb4e7b3f282423d553a21b9fdf24f097fe5f

ramsey:
Yes. This needs to be on the 4.x branch.

> I also think the old moontoast/math could be removed I could not find any usage of it 

I think I’m still using moontoast/math in some tests. If not, it can go away.

alexander-schranz:
You are right `Moontoast\Math\BigNumber` is stilled used in the tests. Update could find a way to remove it: https://github.com/ramsey/uuid/pull/426

GrahamCampbell:
I've provided a PR to the v4 branch: https://github.com/ramsey/uuid/pull/434.

PR #468, opened 2022-10-10
SystemNodeProvider: Dismiss zero'd mac address

zedar187:
Dismiss zero'd mac addresses while generating a node.

## Description
This PR adds a foreach and cycles through the fetched interfaces, until a usable one is found. If there is no match, it will keep generating a new random node.

## Motivation and context
The vpn interface on the server is returning a mac address like '00-00-00-00-00-00' thus having a Node filled with zeros. Refs ramsey/uuid-doctrine#186

## How has this been tested?
Comparing the result on the local dev environment (mac) and the production environment (debian). Generate a `$uuid = Uuid::uuid1();` and var_dump the output. To further pinpoint the issue, I've set multiple var_dumps to get to the correct file and method.

## Types of changes

## PR checklist

I haven't worked with tests before, but from the execution logic I guess they shouldn't need adjustment. When executing `composer test` I've got some exceptions in this file (`PHPUnit\Framework\Exception: PHP Fatal error:  Uncaught Exception: Serialization of 'Closure' is not allowed in Standard input code:114`) - but this error is thrown with as well as without my change. As I stated before, I'm not used to tests...

### DIFF
--- a/src/Provider/Node/SystemNodeProvider.php
+++ b/src/Provider/Node/SystemNodeProvider.php
@@ -135,7 +135,11 @@ class SystemNodeProvider implements NodeProviderInterface
 
         $node = '';
         if (preg_match_all(self::IFCONFIG_PATTERN, $ifconfig, $matches, PREG_PATTERN_ORDER)) {
-            $node = $matches[1][0] ?? '';
+            foreach ($matches[1] as $iface) {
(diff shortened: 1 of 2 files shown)


ramsey (commented):
Can you provide a test for this?

ramsey (approved):

PR #472, opened 2022-10-31
Allow passing null value to Uuid validators

jakobmerrild:
## Description
Upgrading from 3.x to 4.x introduces a breaking change in the api of `Uuid::isValid` in that it will no longer accept `null` as input. 

With these changes `null` is now once again accepted as input to `Uuid::isValid` with the obvious result `false`.
## Motivation and context
We use your library extensively to check that various `string|null` values are valid uuids in our code base. While upgrading to `4.x` we noticed that some of these calls, namely those passing in `null`, would fail at runtime.

## How has this been tested?
I have not tested this beyond the additional test input in the relevant `XValidatorTest` files.

## Types of changes

## PR checklist

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -77,6 +77,7 @@ jobs:
         php-version:
           - "8.0"
           - "8.1"
+          - "8.2"
 
     steps:
       - name: "Checkout repository"
@@ -144,6 +145,7 @@ jobs:
         php-version:
           - "8.0"
           - "8.1"
+          - "8.2"
         operating-system:
           - "ubuntu-latest"
           - "windows-latest"
--- a/.github/workflows/merge-dependabot-upgrades.yml
+++ b/.github/workflows/merge-dependabot-upgrades.yml
@@ -17,7 +17,7 @@ jobs:
     steps:
       - name: Auto-Merge
         if: ${{ github.event.workflow_run.conclusion == 'success' }}
-        uses: ridedott/merge-me-action@v2.10.31
+        uses: ridedott/merge-me-action@v2.10.39
         with:
           # This must be used as GitHub Actions token does not support pushing
           # to protected branches.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,23 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+## 4.6.0 - 2022-11-05
+
+### Added
+
+* Add support for version 8, Unix Epoch time UUIDs, as defined in
+  [draft-ietf-uuidrev-rfc4122bis-00, section 5.8][version8]. While still an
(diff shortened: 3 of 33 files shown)


ramsey:
This changes the validator interface and introduces a BC break, so this would require a version 5.0 release.

Is there a reason you need to pass `null`? Can you check for `null` values before calling the validator?

jakobmerrild:
We can of course check for `null` prior to passing values to `Uuid::isValid`, but I didn't think this was a BC break since `string|null` is a super type of `string`. Am I wrong in this?

ramsey:
Changing the interface results in a fatal error like this:

```
Fatal error: Declaration of Foo::validate(string $uuid): bool must be compatible with
ValidatorInterface::validate(?string $uuid): bool in /in/Np3bf on line 10
```

See https://3v4l.org/Np3bf

jakobmerrild:
Of course! I didn't think of that 🙈.
I'm not familiar enough with PHPs type system to know for sure, but I assume that it isn't strong enough to accept `string|null` in the concrete class if the interface specifies only `string`?

ramsey:
We can give it a try... and it appears to work. 😄 

https://3v4l.org/lP0YG

jakobmerrild:
I realised that `Uuid` does not seem to implement `ValidatorInterface`, so it's sufficient to simply move the null check to the static method `::isValid` in order to achieve the same interface as it had in version 3.

jakobmerrild:
@ramsey Have you had a chance to look at the updated changes?

jakobmerrild:
@ramsey Just following up here. Wondering if this is something you'd be willing to include in the project?

ramsey:
I'm a little concerned about the change, since the Uuid class isn't `final`. Others might still extend Uuid, and if they reimplement `isValid()`, then they'll get a fatal error.

For example, see here: https://3v4l.org/SsOGK

jakobmerrild:
I didn't think of that either 🤔 .

Well it's definitely not worth introducing a breaking change in order to revert a breaking change. After all going from version 3 to 4 one would expect breaking changes 🙂 

I had a peak at the changelog. It doesn't seem to mention breaking changes in new major versions. Do you keep a list of these somewhere? Would then simply add to it that the `Uuid::isValid` no longer accepts `null` in version 4

ramsey:
What happened is that the docblock type for this method was always `string` and never `string | null`. In this commit, @garak added all the type hints according to the docblocks: https://github.com/ramsey/uuid/commit/5459113b816f4b1e15d30dffdc1d0e85bdaf4585#diff-898c45cfd882f04a4877c82e14ce28181f8ea6062654772bfb9fc474014bd6a0R686

In version 3.x, there are no type hints on the method, so it allowed anything, but the docblock specifies that it allows only `string`, so it was a bug that we accepted `null`, and this bug was fixed in the first alpha version of 4.0.0.

I'm sorry for the confusion.

jakobmerrild:
I understand. I guess we'll have to clean up our code base at some point 🙂 . For now we're pinning your library to `3.9`. 
Thanks for your time.

ramsey:
Sorry for the hassle. I just released 3.9.7, which fixes a few things so the library has no problem running on PHP 8.1 or 8.2.

PR #479, opened 2022-12-15
Bump minimum version of ramsey/collection

driesvints:
By bumping ramsey/collection to the latest version that supports PHP 8.1, we prevent builds from third party libraries from failing when they install dependencies with `prefer-lowest`. 

## Description
If right now, a library requires `ramsey/uuid` it'll install the lowest version of ramsey/collection of its ^1.0 constraint on PHP 8.1 instead of the actual compatible one of v1.2.0. By bumping the minimum version we make sure it'll install the latest version when a third party tries to test its lowest bounds with `prefer-lowest`. 

## Motivation and context
We'll be able to remove the workaround here: https://github.com/laravel/framework/blob/9.x/.github/workflows/tests.yml#L76

## How has this been tested?
We'll be able to test this once a new version of ramsey/uuid is tagged.

## Types of changes

## PR checklist

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
         "php": "^8.0",
         "ext-json": "*",
         "brick/math": "^0.8.8 || ^0.9 || ^0.10",
-        "ramsey/collection": "^1.1.2"
+        "ramsey/collection": "^1.2"
     },
     "require-dev": {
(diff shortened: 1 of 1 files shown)


ramsey:
Thank you for contributing! 🎉

PR #486, opened 2023-01-10
fix: Amends psalm assertion syntax on `Uuid::isValid()` to prevent incorrect type inference

### LINKED ISSUE #478: Misused Psalm Assertion for isValid() method?
## Do I misunderstand the @psalm-assert-if-true Assertion?

The `Uuid::isValid` method has a psalm assertion `@psalm-assert-if-true non-empty-string $uuid` set. 
This assertion seems to lead Psalm to think that only non-empty-strings can be given to this method and any false validation means that the given string is empty.  
At least that's what I'm guessing after receiving the following error 
`All possible types for this argument were invalidated - This may be dead code`
in a project of mine.

## Example code

Here's a simplified version of the problematic code. I've just mocked a Uuid::isValid implementation for demo purposes.

https://psalm.dev/r/764e8e9309

gsteel:
## Description

The reason for this change is described in #478

Closes #478

### DIFF
--- a/src/Uuid.php
+++ b/src/Uuid.php
@@ -569,7 +569,7 @@ class Uuid implements UuidInterface
      * @psalm-pure note: changing the internal factory is an edge case not covered by purity invariants,
      *             but under constant factory setups, this method operates in functionally pure manners
      *
-     * @psalm-assert-if-true non-empty-string $uuid
+     * @psalm-assert-if-true =non-empty-string $uuid
      */
     public static function isValid(string $uuid): bool
     {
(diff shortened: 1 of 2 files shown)


ramsey (approved):

gsteel:
Just realised that I should be able to remove the PHPStan suppression shipped in the original PR - I'll rebase and try it…

gsteel:
The previous error suppression will have to stay for now :(

ramsey:
I'm not sure why the workflow didn't run when you did the force push. Normally, that should have triggered a build.

gsteel:
I've pushed an additional test case to trigger CI. GHA was being weird for me on other repos earlier!

ramsey:
Thank you for contributing! 🎉

PR #491, opened 2023-02-06
Add explicit `Stringable` interface to `UuidInterface`

boesing:
## Description
Even tho, the `Stringable` interface is implicit on all objects containing the method `__toString`, it is recommended by PHP to explicitly setting that interface.

From php.net documentation:
> [...] Unlike most interfaces, **Stringable** is implicitly present on any class that has the magic `__toString()` method defined, although it can and should be declared explicitly.

## Motivation and context

I've used `Stringable` notation recently in an upstream project and then realized that the `UuidInterface` does not explicitly implements the interface. Due to the magic PHP provides starting with PHP 8.0, it still implements `Stringable` and thus I was able to use it but to avoid future issues with i.e. PHP 9 where this behavior might change, I wanted to add this here.

## How has this been tested?

I've added 2 new checks to the existing tests. 
Please note that these tests won't fail even if we remove the interface again due to the fact that PHP adds it implicitly whenever a `__toString` method is available.

## Types of changes

## PR checklist

- [X] I have read the **CONTRIBUTING.md** document.
- [X] I have added tests to cover my changes.

### DIFF
--- a/src/UuidInterface.php
+++ b/src/UuidInterface.php
@@ -19,6 +19,7 @@ use Ramsey\Uuid\Fields\FieldsInterface;
 use Ramsey\Uuid\Type\Hexadecimal;
 use Ramsey\Uuid\Type\Integer as IntegerObject;
 use Serializable;
+use Stringable;
 
 /**
  * A UUID is a universally unique identifier adhering to an agreed-upon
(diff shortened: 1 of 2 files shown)


ramsey:
I'm not sure why phpcs is failing the build for warnings. That's odd. Can you break those assertions across multiple lines so they don't go over 120 characters? In the meantime, I'll figure out how to tell PHP CodeSniffer that warnings aren't failures. 😄

PR #588, opened 2025-02-19
update RoundingMode to be compatiable with already upgraded brick/math

### LINKED ISSUE #586: Outdated version of RoundingMode.php results in fatal
This package currently uses older version of  brick/math

https://github.com/ramsey/uuid/blob/4.x/src/Math/RoundingMode.php

Latest version: https://github.com/brick/math/blob/master/src/RoundingMode.php

Hence, if there are any other plugins that use latest `brick/math` it results in fatal error. This is mainly due to switch to enums.

kurund:
Fixes https://github.com/ramsey/uuid/issues/586


ramsey:
This seems like a BC break. I don't think I can include this in the current 4.x series without potentially breaking a lot of users' applications.

ramsey:
@kurund, thanks for your contribution!

After more investigation, I don't think there's a need for this change, since this array performs the translation between this library's `RoundingMode` and brick/math's `RoundingMode`:

https://github.com/ramsey/uuid/blob/7eec9c220262d163b175f0537454795b2bf2bc19/src/Math/BrickMathCalculator.php#L34-L45

That translation allows code like this to work on both older versions of brick/math and newer versions that have the enum:

https://github.com/ramsey/uuid/blob/7eec9c220262d163b175f0537454795b2bf2bc19/src/Math/BrickMathCalculator.php#L80-L99

If we change this library's `RoundingMode` to be an enum, then this library will no longer be compatible with versions of brick/math earlier than `0.12`.

PR #596, opened 2025-06-06
remove useless ext-json requirement

### LINKED ISSUE #595: ext-json requirement non needed
Since this library has dropped support for PHP 7, the ext-json requirement is redundant (being the extension included in the core since PHP 8.0)

garak:
Fix #595


ramsey:
Thank you for contributing! 🎉

PR #603, opened 2025-06-22
Re-add the @pure annotations

ramsey:
## Description

Add `@pure` annotations where the `@psalm-pure` annotations used to be.

## Motivation and context

These were removed in https://github.com/ramsey/uuid/commit/691c2c816ea574eab667268eca85a1d9046ea599 but should remain in the code base.

For more information, see discussion at https://github.com/ramsey/uuid/commit/691c2c816ea574eab667268eca85a1d9046ea599#r160296998

## How has this been tested?

~Static analysis is now failing on each `@pure` method. I will need some help resolving the failures.~

Thanks to @fillerwriter (see #605), static analysis is now passing!

## Types of changes

## PR checklist

### DIFF
--- a/src/Codec/GuidStringCodec.php
+++ b/src/Codec/GuidStringCodec.php
@@ -32,6 +32,7 @@ class GuidStringCodec extends StringCodec
 {
     public function encode(UuidInterface $uuid): string
     {
+        /** @phpstan-ignore possiblyImpure.methodCall */
         $hex = bin2hex($uuid->getFields()->getBytes());
 
         /** @var non-empty-string */
@@ -52,8 +53,10 @@ class GuidStringCodec extends StringCodec
 
     public function decode(string $encodedUuid): UuidInterface
     {
+        /** @phpstan-ignore possiblyImpure.methodCall */
         $bytes = $this->getBytes($encodedUuid);
 
+        /** @phpstan-ignore possiblyImpure.methodCall, possiblyImpure.methodCall */
         return $this->getBuilder()->build($this, $this->swapBytes($bytes));
     }
 
--- a/src/Codec/OrderedTimeCodec.php
+++ b/src/Codec/OrderedTimeCodec.php
@@ -87,12 +87,10 @@ class OrderedTimeCodec extends StringCodec
 
         $uuid = parent::decodeBytes($rearrangedBytes);
 
-        if (
-            /** @phpstan-ignore possiblyImpure.methodCall */
-            !($uuid->getFields() instanceof Rfc4122FieldsInterface)
-            /** @phpstan-ignore possiblyImpure.methodCall */
-            || $uuid->getFields()->getVersion() !== Uuid::UUID_TYPE_TIME
-        ) {
+        /** @phpstan-ignore possiblyImpure.methodCall */
+        $fields = $uuid->getFields();
+
+        if (!$fields instanceof Rfc4122FieldsInterface || $fields->getVersion() !== Uuid::UUID_TYPE_TIME) {
             throw new UnsupportedOperationException(
                 'Attempting to decode a non-time-based UUID using OrderedTimeCodec',
             );
--- a/src/Codec/OrderedTimeCodec.php
+++ b/src/Codec/OrderedTimeCodec.php
@@ -51,12 +51,15 @@ class OrderedTimeCodec extends StringCodec
     public function encodeBinary(UuidInterface $uuid): string
     {
         if (
+            /** @phpstan-ignore possiblyImpure.methodCall */
             !($uuid->getFields() instanceof Rfc4122FieldsInterface)
+            /** @phpstan-ignore possiblyImpure.methodCall */
             || $uuid->getFields()->getVersion() !== Uuid::UUID_TYPE_TIME
         ) {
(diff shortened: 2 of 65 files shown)


ramsey:
@Ocramius I've re-added the `@pure` annotations, so static analysis is now failing (as expected, I guess). I could use some direction on what's needed to fix the failures. Thanks!

fillerwriter:
👋 Hello! I've spent some time on this branch, and here's my current thinking, along with an approach.

- First principles: As we all know, [a pure function is a function that will return the same value every time it's called with the same inputs, and has no side effects in the codebase](https://en.wikipedia.org/wiki/Pure_function).
- First principles: `ramsey/uuid` also features a factory subsystem that allows for pluggable uuid generation, which is used internally for automated testing purposes. In it's current state, this factory subsystem uses static variables to change how it creates UUID's, which by definition makes anything it touches an impure function. With normal usage, this subsystem doesn't change, and based on the [codebases's internal comments](https://github.com/ramsey/uuid/commit/691c2c816ea574eab667268eca85a1d9046ea599#diff-898c45cfd882f04a4877c82e14ce28181f8ea6062654772bfb9fc474014bd6a0L560), can be considered pure with normal usage.
- **Some of the methods we label as `@pure\@psalm-pure` are legitimately pure functions, and should keep the `@pure` label.** For example, [the comment on `BinaryUtils`](https://github.com/ramsey/uuid/commit/691c2c816ea574eab667268eca85a1d9046ea599#r160296998) from the original PR is a legitimate pure function. It's also passing the current `phpstan` scan as is.
- The code where `phpstan` is currently failing are largely methods either directly or indirectly touching this factory system. In earlier versions of the codebase, [there were `@psalm-ignore` directives that suppressed errors](https://github.com/ramsey/uuid/commit/691c2c816ea574eab667268eca85a1d9046ea599#diff-898c45cfd882f04a4877c82e14ce28181f8ea6062654772bfb9fc474014bd6a0L560). **We should add `php-stan` equivalent ignore directives with comparable notes.** 

I'm planning on working through the error list this morning, but wanted to drop this into the PR thread to make sure that folks were on board with this approach before getting too far into it.

ramsey:
Thanks so much for your help, @fillerwriter!

@Ocramius, I'd love for you to take a final glance at this before I merge it into the `4.x` branch. We're now adding `@pure` in some places where it wasn't before, and I want to make sure this doesn't also create a BC break.

If it's considered new, backward-compatible functionality, then all is well, and I'll tag a 4.x.0 release. Otherwise, I want to remove the annotations that cause BC breaks for `4.x`, while keeping everything for the `5.x` branch.

ramsey:
@fillerwriter, I addressed your concerns about the PHPStan errors in the tests here: https://github.com/ramsey/uuid/pull/603/commits/198a34a8b4c11319610bbeb035b87831cfae2764

I'm still not sure why it only just now started complaining about that.

fillerwriter:
Nice! Re: PHPStan errors in tests, that's a nice trick! I didn't know you could filter errors based on the folder.

ramsey:
> I didn't know you could filter errors based on the folder.

I didn't, either, but I suspected you could, so I figured that out tonight. 🙂

Ocramius:
@ramsey fwiw, @fillerwriter's analysis is correct, and the suppressions correspond to my initial design too (which is also what highlighted the entire factory problematic under the hood)

src/Codec/GuidStringCodec.php
```diff
@@ -32,6 +32,7 @@ class GuidStringCodec extends StringCodec
 {
     public function encode(UuidInterface $uuid): string
     {
+        /** @phpstan-ignore possiblyImpure.methodCall */
```
Ocramius, line 35: Deep inside these implementations, I'd suggest suppressing the entire category of errors at class level, where needed

src/Codec/OrderedTimeCodec.php
```diff
@@ -85,7 +88,9 @@ public function decodeBytes(string $bytes): UuidInterface
         $uuid = parent::decodeBytes($rearrangedBytes);
 
         if (
+            /** @phpstan-ignore possiblyImpure.methodCall */
             !($uuid->getFields() instanceof Rfc4122FieldsInterface)
+            /** @phpstan-ignore possiblyImpure.methodCall */
             || $uuid->getFields()->getVersion() !== Uuid::UUID_TYPE_TIME
```
Ocramius: Interesting here: `getFields()` gets called twice, which is a small thing to optimize away, and reduces us to a single suppression

Ocramius (approved):
From my PoV, this patch reflects the initial design when `@pure` was introduce.

The only thing that is a bit annoying is the fact that we need to suppress inline so much, but on the other end, it was kinda expected.

LGTM

src/Codec/GuidStringCodec.php
ramsey: I tried adding `@phpstan-ignore` at the method level, and it wouldn't work. PHPStan was telling me there wasn't a `possiblyImpure.methodCall` error on the next line. The only way I could find to squelch the errors is to inline them like this or add them to the config file, but I'd rather leave them in the code.

PR #605, opened 2025-06-24
Fix pure annotations - First pass

fillerwriter:
Coming from [Re-add the @pure annotations PR](https://github.com/ramsey/uuid/pull/603), this is an attempt to fix the errors raised by the current phpstan settings. 

## Description
I went through each of the errors raised by phpstan with the following approach.

- If a method is part of an `@immutable` class, we can consider it pure, assuming it only affects internal variables.
- If a potentially pure method is calling a class's method that is only swapped during testing (and not during normal usage), then we can consider the calling method pure.
- If a class is marked deprecated, don't bother with attempting to mark it pure or immutable.

There's _a lot_ of ignores that have been placed for this PR, so a second set of eyes is welcome and appreciated.

@ramsey, I have this PR set to merge into `uuid/fix-pure-annotations`, but happy to switch to `uuid/main` if that makes more sense.

One thing to note that bugs me about the initial state of this PR is `phpstan` flags about 40 different `method.resultUnused` issues in the tests folder. Generally, these tests are either for testing exceptions, or as part of a benchmarking suite. In both cases, the code works as intended. However, having to explicitly flag `phpstan` to ignore these feels like a downgrade to the developer experience here, and might be an additional minor hindrance to external contrib. My view is likely skewed by placing all the exceptions at the same time, but I would be open to a different approach that automatically filters those warnings in the test folder.

## How has this been tested?
This has been tested by running `composer run-script test` locally. No actual code changes were made, this should all be `phpstan` comments.

## Types of changes

## PR checklist

### DIFF
--- a/src/Codec/GuidStringCodec.php
+++ b/src/Codec/GuidStringCodec.php
@@ -32,6 +32,7 @@ class GuidStringCodec extends StringCodec
 {
     public function encode(UuidInterface $uuid): string
     {
+        /** @phpstan-ignore possiblyImpure.methodCall */
         $hex = bin2hex($uuid->getFields()->getBytes());
 
         /** @var non-empty-string */
@@ -52,8 +53,10 @@ class GuidStringCodec extends StringCodec
 
     public function decode(string $encodedUuid): UuidInterface
     {
+        /** @phpstan-ignore possiblyImpure.methodCall */
         $bytes = $this->getBytes($encodedUuid);
 
+        /** @phpstan-ignore possiblyImpure.methodCall, possiblyImpure.methodCall */
         return $this->getBuilder()->build($this, $this->swapBytes($bytes));
     }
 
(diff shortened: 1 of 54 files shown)


ramsey:
Quick note before I dive into reviewing this...

> One thing to note that bugs me about the initial state of this PR is phpstan flags about 40 different method.resultUnused issues in the tests folder. 

I know what's going on here, but it's strange to me that PHPStan only started complaining about these when I added the `@pure` annotations back to the source. Very weird.

src/Codec/GuidStringCodec.php
```diff
@@ -52,8 +53,10 @@ public function encode(UuidInterface $uuid): string
 
     public function decode(string $encodedUuid): UuidInterface
     {
+        /** @phpstan-ignore possiblyImpure.methodCall */
         $bytes = $this->getBytes($encodedUuid);
 
+        /** @phpstan-ignore possiblyImpure.methodCall, possiblyImpure.methodCall */
```
ramsey, line 59: It's so wild you have to list the error ID twice, since there are two occurrences of it here. It makes sense, but I wouldn't have guessed it.

ramsey (approved):
Thank you so much! This looks great!

PR #633, opened 2026-03-12
Allow brick math 0.15 and 0.16

BackEndTea:
0.15 removed the 'old' rounding mode constants, so this includes a BC layer in the `BrickMathCalculator`

## Description
This MR allows installation of brick math versions 0.15 and 0.16.

0.14 changed the rounding mode to an enum, but kept backwards compatability. in 0.15 this BC layer was removed. 

To support older brick/math versions (and older PHP versions), we introduce a BC layer in the `BrickMathCalculator` class.

## How has this been tested?
I have ran the phpunit tests with brick/math 0.14, 0.15 and 0.16.

## Types of changes

## PR checklist

PHPUnit test are working for this MR, but there are some PHPStan failures. I'm not sure how to move forward with those.


balu-lt:
After creating PR #634 for the `5.x` branch, I noticed you had already opened this PR for the `4.x` branch.

By the way, is there an ETA for the v5 release?

tests/Builder/FallbackBuilderTest.php
```diff
@@ -25,41 +21,41 @@
...
-        $fallbackBuilder = new FallbackBuilder([$builder1, $builder2, $builder3]);
-
-        $this->expectException(BuilderNotFoundException::class);
-        $this->expectExceptionMessage(
-            'Could not find a suitable builder for the provided codec and fields'
-        );
-
-        $fallbackBuilder->build($codec, $bytes);
-    }
+//    public function testBuildThrowsExceptionAfterAllConfiguredBuildersHaveErrored(): void
```
ramsey: Why are these tests commented out?

ramsey:
It looks like brick/math broke a lot of BC with earlier releases and removed or moved a lot of things. That's why there are so many PHPStan errors. Is that something you're able to help clean up?

tests/Builder/FallbackBuilderTest.php
BackEndTea: My bad, i think i commented them out to debug something. It should be fine to uncomment them again

BackEndTea:
> It looks like brick/math broke a lot of BC with earlier releases and removed or moved a lot of things. That's why there are so many PHPStan errors. Is that something you're able to help clean up?

No problem. My suggestion would be to 'simply' add a bunch of `@phpstan-ignore` comments to the errors that are due to the BC layer.

I think the most maintainable solution for this library is to get this MR (or another with a BC layer) merged and tagged. 
Then in the next tag remove the BC layer again, and require a higher version of brick/match. 

This would give users of the package an easier upgrade path, as this can be upgraded first, and then brick/math later. But it would introduce the least strain on maintainability.

Is that a solution that works for you?

BackEndTea:
I have added the ignores, and made an example of dropping the BC layer again and requiring the newer brick math versions: https://github.com/BackEndTea/uuid/pull/1

src/Math/BrickMathCalculator.php
```diff
@@ -31,18 +31,11 @@
  */
 final class BrickMathCalculator implements CalculatorInterface
 {
-    private const ROUNDING_MODE_MAP = [
```
talkinnl, line 34:
```php
    private const ROUNDING_MODE_MAP = [
        RoundingMode::UNNECESSARY => BrickMathRounding::Unnecessary ?? BrickMathRounding::UNNECESSARY,
        ...
```

Alternatively and a matter of taste, but this could also work instead of the static $roundingModeMap.
With this approach, you don't need `defined()`. The whole compat layer would be these 10 lines.

jnoordsij:
With https://github.com/ramsey/uuid/pull/638 merged I guess this is now obsolete?

Issue #636, opened 2026-03-20
Allow installing `brick/math:~0.17.0`


localheinz:
Several pull requests have been opened to allow installing newer versions of `brick/math`:

- #632 
- #633 
- #634 

Not allowing newer versions of `brick/math` blocks projects using

- `ramsey/uuid`
- `brick/math`
- `brick/money`

and wishing to update to the latest versions of `brick/math` and `brick/money`. 

An update of `brick/math` and `brick/money` is only possible with a downgrade of `ramsey/uuid` to `ramsey/uuid` to a version that issues deprecations on PHP 8.4:

```
Ramsey\Uuid\UuidFactory::__construct(): Implicitly marking parameter $features as nullable is deprecated, the explicit nullable type must be used instead
```

### Releases of `brick/math` that we can't update to

[`ramsey/uuid:4.9.2`](https://github.com/ramsey/uuid/blob/4.9.2/composer.json#L13) prevents updating to

- [`brick/math:0.15.0`](https://github.com/brick/math/releases/tag/0.15.0) (released on 2026-02-20)
- [`brick/math:0.16.0`](https://github.com/brick/math/releases/tag/0.16.0) (released on 2026-03-06)
- [`brick/math:0.16.1`](https://github.com/brick/math/releases/tag/0.16.1) (released on 2026-03-09)
- [`brick/math:0.16.2`](https://github.com/brick/math/releases/tag/0.16.2) (released on 2026-03-15)
- [`brick/math:0.17.0`](https://github.com/brick/math/releases/tag/0.17.0) (released on 2026-03-17)
- [`brick/math:0.17.1`](https://github.com/brick/math/releases/tag/0.17.1) (released on 2026-04-19)

### Releases of `brick/money` that we can't update to

[`ramsey/uuid:4.9.2`](https://github.com/ramsey/uuid/blob/4.9.2/composer.json#L13) prevents updating to

- [`brick/money:0.12.0`](https://github.com/brick/money/releases/tag/0.12.0) (released on 2026-03-03)
- [`brick/money:0.12.1`](https://github.com/brick/money/releases/tag/0.12.1) (released on 2026-03-11)
- [`brick/money:0.12.2`](https://github.com/brick/money/releases/tag/0.12.2) (released on 2026-03-19)
- [`brick/money:0.12.3`](https://github.com/brick/money/releases/tag/0.12.3) (released on 2026-03-23)
[...]

PR #638, opened 2026-04-23
[4.x] Upgrade `brick/math` to support versions ^0.14–^0.17

delolmo:
## Summary

I noticed there have been a few attempts at adding support for newer brick/math versions, so I wanted to take a stab at it with an approach that keeps full backward compatibility.

The main challenge is that brick/math 0.15 renamed the `RoundingMode` constants from `UPPER_SNAKE_CASE` to `PascalCase` and changed `MathException` from a class to an interface. Rather than dropping support for older versions, I introduced a small internal `BrickMathRoundingMode` class that resolves the correct constant at runtime using `defined()` with a fallback. It ain't pretty, but it works seamlessly across all versions.

I also tightened the PHPDoc types on `CalculatorInterface` to satisfy the stricter type annotations introduced in newer brick/math releases, and adjusted a couple of test assertions where error message wording changed.

The full `composer test` suite (lint, benchmarks, PHPStan, unit tests) passes on both sides of the 0.15 breaking-change boundary, and unit tests pass on every supported version: 0.8.16, 0.9, 0.10, 0.11, 0.12, 0.13, 0.14, 0.15, 0.16, and 0.17.

See #636.

## Test plan

- [x] `composer test` passes with brick/math 0.14
- [x] `composer test` passes with brick/math 0.17
- [x] Unit tests pass across all 10 supported major versions (0.8.16 through 0.17)
- [x] CI passes with both `locked` and `highest` dependency strategies


ramsey (approved):

miken32:
@ramsey this still needs a new release pushed out before it shows up on Packagist

ahsankhatri:
@ramsey Thanks for merging this PR. Eagerly waiting for a new release :)

aletundo:
@ramsey we are all on the same boat :D Any news for the release?

neoighodaro:
Just checking in like the rest. Thanks in advance.

PR #642, opened 2026-08-05
Fix: Add support for `brick/math:^0.19`, `brick/math:^0.20`, and `brick/math:^1.0`

localheinz:
This pull request

- [x] adds support for `brick/math:^0.19`, `brick/math:^0.20`, and `brick/math:^1.0`

❗Blocked by #643.

Fixes #644.

💁‍♂️ Note that the branch for this pull request contains a cherry-pick of d8629c163cea5b498a5f174a300727e201f6736c from #643.

PR #643, opened 2026-08-05
Fix: Use `foreach` instead of `array_map()` in `UuidStringConversionBench`

localheinz:
This pull request

- [x] uses `foreach` instead of `array_map()` in `UuidStringConversionBench`

❗ Blocks #642.

💁‍♂️ Maybe it would be better to run the

- `coding-standards`
- `static-analysis`

with locked dependencies instead?


ghostwriter (approved):

localheinz:
Closing as it has been applied in https://github.com/ramsey/uuid/pull/642.

PR #646, opened 2026-09-13
Fix: Downgrade `brick/math` for locked dependencies on PHP 8.0 and 8.1 only

localheinz:
This pull request

- [x] downgrades `brick/math` for locked dependencies on PHP 8.0 and 8.1 only


.github/workflows/continuous-integration.yml
```diff
@@ -197,6 +197,7 @@ jobs:
           ini-values: "memory_limit=-1"
 
       - name: "Downgrade brick/math in lock file for PHP 8.0"
+        if: ${{ matrix.dependency-versions == 'locked' && matrix.php-version == '8.0' }}
```
localheinz: Note that bec7176 missed adding a condition here.

ramsey: Whoops!

ramsey:
Thank you for contributing! 🎉

localheinz:
Thank you, @ramsey!

PR #649, opened 2026-09-19
Migration from Mockery to Double

jasonmccreary:
## Description
This PR migrations the test suite from Mockery to [Double](https://testdoublephp.com). Double is a modern PHP testing library that focuses on developer experience.

All changes are only to the test suite and development dependencies.

## Motivation and context
This is primary an evaluation of Double against a real, mock-heavy suite as a Mockery replacement. However, along the way it revealed many of the common misuses of Mockery within this test suite. So I decided to submit the PR.

## How has this been tested?
PHPUnit. PHPStan and PHPCS are clean. `PeclUuid*` tests need the `uuid` extension, so they're pending CI.

## Types of changes
Beyond the conversion itself:
- Four bare `shouldReceive()` calls were strengthened to `expects()`.
- A dead `currentTime` stub on `TimeProviderInterface` was removed.
- `isMax()` is stubbed via a test-only and `FieldsInterface` doesn't declare it.
- `PHPMockery` was replaced with `php-mock/php-mock`.
- The "factory never constructed" check in `RandomLibAdapterTest` was replaced with a check that the injected generator is used.

## PR checklist
  - `docs/testing.rst` still shows `Mockery::mock()` examples.

*Note:* Double requires a supported version of PHP. So technically it requires 8.2 or higher. As a `require-dev` dependency this has no real-world impact on the package compatibility. But maybe this should target `5.x`...


jasonmccreary:
Upon closer inspection. This should likely target 5.x given its changes and the PHP 8.2 requirement.

